PR #1, opened 2024-07-03
Blender: Expose frame range and unit conversion attributes on publish instances

iLLiCiTiT:
## Changelog Description

Improved creators and publish instances to allow per instance frame ranges to be configured instead of using folder-level attributes.

### Additional info

**Note:** A caveat of this is that now the frame ranges of instances aren't 1:1 tied to the scene's frame range but to what is configured on the publish instance in the publisher UI.

Transferred PR from https://github.com/ynput/ayon-core/pull/658

## Testing notes:
The creators and the extractors that have been updated should be tested to check if they're using the information set in the creator.

1. FBX extractors should now have a Unit Scale attribute that works in a sensible manner:
    - [x] These should provide backward compatible results (existing instances should generate the same scale exports when now published with this PR as before without this PR
    - [x] Model publish works, with FBX unit scale providing different scaling
    - [x] Animation publish works, with FBX unit scale providing different scaling

2. All Extractors should now adhere to a frame range configured on the instance instead of exporting all.
    - [x] FBX export should not have animation data outside of the frame range.
    - [x] Animation export should not have animation data outside of the frame range.
    - [x] Review export should match the configured frame range on the instance.

### DIFF
--- a/client/ayon_blender/plugins/create/create_camera.py
+++ b/client/ayon_blender/plugins/create/create_camera.py
@@ -2,6 +2,7 @@
 
 import bpy
 
+from ayon_core.lib import NumberDef
 from ayon_blender.api import plugin, lib
 from ayon_blender.api.pipeline import AVALON_INSTANCES
 
@@ -40,3 +41,15 @@ class CreateCamera(plugin.BlenderCreator):
             camera_obj.parent = asset_group
 
         return asset_group
+
+    def get_instance_attr_defs(self):
+        defs = lib.collect_animation_defs()
+
+        defs.append(
+            NumberDef("unitScale",
+                      label="Unit Scale (FBX)",
+                      default=1.0,
+                      tooltip="Scale of the model, valid only for FBX export.")
+        )
+
+        return defs
--- a/client/ayon_blender/plugins/create/create_camera.py
+++ b/client/ayon_blender/plugins/create/create_camera.py
@@ -2,7 +2,6 @@
 
 import bpy
 
-from ayon_core.lib import NumberDef
 from ayon_blender.api import plugin, lib
 from ayon_blender.api.pipeline import AVALON_INSTANCES
 
@@ -44,12 +43,4 @@ class CreateCamera(plugin.BlenderCreator):
 
     def get_instance_attr_defs(self):
         defs = lib.collect_animation_defs()
-
-        defs.append(
-            NumberDef("unitScale",
-                      label="Unit Scale (FBX)",
-                      default=1.0,
-                      tooltip="Scale of the model, valid only for FBX export.")
-        )
-
         return defs
--- a/client/ayon_blender/plugins/publish/extract_abc.py
+++ b/client/ayon_blender/plugins/publish/extract_abc.py
@@ -44,6 +44,20 @@ class ExtractABC(plugin.BlenderExtractor, publish.OptionalPyblishPluginMixin):
         context = plugin.create_blender_context(
             active=asset_group, selected=selected)
 
+        scene = bpy.context.scene
+        frame_start = scene.frame_start
+        frame_end = scene.frame_end
+        frame_step = scene.frame_step
+        fps = scene.render.fps
+        fps_base = scene.render.fps_base
+        scene.frame_start = instance.data.get("frameStart", frame_start)
+        scene.frame_end = instance.data.get("frameEnd", frame_end)
+        scene.frame_step = instance.data.get("frameStep", frame_step)
+        inst_fps = instance.data.get("fps")
+        if inst_fps:
+            scene.render.fps = inst_fps
+            scene.render.fps_base = 1
+
         with bpy.context.temp_override(**context):
             # We export the abc
             bpy.ops.wm.alembic_export(
@@ -52,6 +66,12 @@ class ExtractABC(plugin.BlenderExtractor, publish.OptionalPyblishPluginMixin):
                 flatten=False,
                 subdiv_schema=attr_values.get("subdiv_schema", False)
             )
+
+        scene.frame_start = frame_start
+        scene.frame_end = frame_end
+        scene.frame_step = frame_step
+        scene.render.fps = fps
+        scene.render.fps_base = fps_base
 
         plugin.deselect_all()
 
--- a/client/ayon_blender/plugins/publish/extract_abc.py
+++ b/client/ayon_blender/plugins/publish/extract_abc.py
@@ -4,7 +4,7 @@ import bpy
 
 from ayon_core.lib import BoolDef
 from ayon_core.pipeline import publish
-from ayon_blender.api import plugin
+from ayon_blender.api import plugin, lib
 
 
 class ExtractABC(plugin.BlenderExtractor, publish.OptionalPyblishPluginMixin):
@@ -44,34 +44,29 @@ class ExtractABC(plugin.BlenderExtractor, publish.OptionalPyblishPluginMixin):
         context = plugin.create_blender_context(
             active=asset_group, selected=selected)
 
-        scene = bpy.context.scene
-        frame_start = scene.frame_start
-        frame_end = scene.frame_end
-        frame_step = scene.frame_step
-        fps = scene.render.fps
-        fps_base = scene.render.fps_base
-        scene.frame_start = instance.data.get("frameStart", frame_start)
-        scene.frame_end = instance.data.get("frameEnd", frame_end)
-        scene.frame_step = instance.data.get("frameStep", frame_step)
-        inst_fps = instance.data.get("fps")
-        if inst_fps:
-            scene.render.fps = inst_fps
-            scene.render.fps_base = 1
+        scene_overrides = {
+            "frame_start": instance.data.get("frameStart"),
+            "frame_end": instance.data.get("frameEnd"),
+            "frame_step": instance.data.get("frameStep"),
+            "render.fps": instance.data.get("fps")
+        }
+        # Skip None value overrides
+        scene_overrides = {
+            key: value for key, value in scene_overrides.items()
+            if value is not None
+        }
+        if "render.fps" in scene_overrides:
+            scene_overrides["render.fps_base"] = 1
 
-        with bpy.context.temp_override(**context):
-            # We export the abc
-            bpy.ops.wm.alembic_export(
-                filepath=filepath,
-                selected=True,
-                flatten=False,
-                subdiv_schema=attr_values.get("subdiv_schema", False)
-            )
-
-        scene.frame_start = frame_start
-        scene.frame_end = frame_end
-        scene.frame_step = frame_step
-        scene.render.fps = fps
-        scene.render.fps_base = fps_base
+        with lib.attribute_overrides(bpy.context.scene, scene_overrides):
+            with bpy.context.temp_override(**context):
+                # We export the abc
+                bpy.ops.wm.alembic_export(
+                    filepath=filepath,
+                    selected=True,
+                    flatten=False,
+                    subdiv_schema=attr_values.get("subdiv_schema", False)
+                )
 
         plugin.deselect_all()
 
--- a/client/ayon_blender/plugins/publish/extract_abc.py
+++ b/client/ayon_blender/plugins/publish/extract_abc.py
@@ -48,7 +48,6 @@ class ExtractABC(plugin.BlenderExtractor, publish.OptionalPyblishPluginMixin):
             "frame_start": instance.data.get("frameStart"),
             "frame_end": instance.data.get("frameEnd"),
             "frame_step": instance.data.get("frameStep"),
-            "render.fps": instance.data.get("fps")
         }
         # Skip None value overrides
         scene_overrides = {
--- a/client/ayon_blender/plugins/publish/extract_abc_animation.py
+++ b/client/ayon_blender/plugins/publish/extract_abc_animation.py
@@ -55,6 +55,20 @@ class ExtractAnimationABC(
         context = plugin.create_blender_context(
             active=asset_group, selected=selected)
 
+        scene = bpy.context.scene
+        frame_start = scene.frame_start
+        frame_end = scene.frame_end
+        frame_step = scene.frame_step
+        fps = scene.render.fps
+        fps_base = scene.render.fps_base
+        scene.frame_start = instance.data.get("frameStart", frame_start)
+        scene.frame_end = instance.data.get("frameEnd", frame_end)
+        scene.frame_step = instance.data.get("frameStep", frame_step)
+        inst_fps = instance.data.get("fps")
+        if inst_fps:
+            scene.render.fps = inst_fps
+            scene.render.fps_base = 1
+
         with bpy.context.temp_override(**context):
             # We export the abc
             bpy.ops.wm.alembic_export(
@@ -62,6 +76,12 @@ class ExtractAnimationABC(
                 selected=True,
                 flatten=False
             )
+
+        scene.frame_start = frame_start
+        scene.frame_end = frame_end
+        scene.frame_step = frame_step
+        scene.render.fps = fps
+        scene.render.fps_base = fps_base
 
         plugin.deselect_all()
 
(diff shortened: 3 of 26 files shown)


moonyuet (approved):
Didn't test, but the code looks good for me

BigRoy:
The PR title does not match at all what the code changes do - Something is wrong? :)

client/ayon_blender/plugins/create/create_camera.py
```diff
@@ -40,3 +41,15 @@ def create(
...
+
+    def get_instance_attr_defs(self):
+        defs = lib.collect_animation_defs()
+
+        defs.append(
+            NumberDef("unitScale",
+                      label="Unit Scale (FBX)",
+                      default=1.0,
+                      tooltip="Scale of the model, valid only for FBX export.")
+        )
```
BigRoy, line 53: If we need to repeat this for _each_ creator - shouldn't this be an attribute definition on the Extractor plugins instead?

BigRoy: Ah I had this commented already here: https://github.com/ynput/ayon-core/pull/658#discussion_r1644003785

client/ayon_blender/plugins/publish/extract_abc.py
```diff
@@ -44,6 +44,20 @@ def process(self, instance):
...
+        frame_step = scene.frame_step
+        fps = scene.render.fps
+        fps_base = scene.render.fps_base
+        scene.frame_start = instance.data.get("frameStart", frame_start)
+        scene.frame_end = instance.data.get("frameEnd", frame_end)
+        scene.frame_step = instance.data.get("frameStep", frame_step)
+        inst_fps = instance.data.get("fps")
+        if inst_fps:
+            scene.render.fps = inst_fps
+            scene.render.fps_base = 1
```
BigRoy, line 59: The [alembic export command](https://docs.blender.org/api/current/bpy.ops.wm.html#bpy.ops.wm.alembic_export) can take a `start` and `end`? Why do we need to change the scene?

Additionally, I would avoid exposing the FPS on the instance for this if it's this tied to the scene. Coming from maya where often changing the FPS has so much influence on the scene that it hardly makes a reproducable output file. But maybe that's different in Blender?

If we do need to change the scene to export the way we want - then we should use a context manager!

Also, if the logic is this involved and we're duplicating it this many times we should make it into a library function.

See comment here: https://github.com/ynput/ayon-core/pull/658#discussion_r1644008494

client/ayon_blender/plugins/publish/extract_camera_fbx.py
```diff
@@ -49,8 +49,22 @@ def process(self, instance):
...
+        frame_end = scene.frame_end
+        frame_step = scene.frame_step
+        fps = scene.render.fps
+        fps_base = scene.render.fps_base
+        scene.unit_settings.scale_length = instance.data.get(
+            "unitScale", scale_length)
+        scene.frame_start = instance.data.get("frameStart", frame_start)
+        scene.frame_end = instance.data.get("frameEnd", frame_end)
+        scene.frame_step = instance.data.get("frameStep", frame_step)
+        inst_fps = instance.data.get("fps")
```
BigRoy, line 64: Shouldn't this just use the `global_scale` argument of the export function? [see here](https://docs.blender.org/api/current/bpy.ops.export_scene.html#module-bpy.ops.export_scene) or why is changing the scene better?

If we are to change the scene - then this should be done via a context manager!

> agreed. This should be done with the contextmanager + context.exitstack()
> You can get some inspirations here: https://github.com/ynput/ayon-core/pull/674#discussion_r1649068696


See: https://github.com/ynput/ayon-core/pull/658#discussion_r1644015067

BigRoy (dismissed):
Copied some comments over from the original PR.

Also wanted to link Libor's test from back then since what he expected to work, doesn't work via this PR: https://github.com/ynput/ayon-core/pull/658#pullrequestreview-2138536723

client/ayon_blender/plugins/publish/extract_camera_fbx.py
BigRoy: The changing of the scene scale length for me makes absolutely no difference on the exported FBX? Also, the "Unit scale" value in the publisher UI is the actual unit scale - and not a _scale factor_ to your scene units... which may be confusing as well.

Why wouldn't we use the `global_scale` instead as scaling factor?

moonyuet:
This is my unit scale setting for blender in AYON(for matching the default unit scale in maya)
image
image
Publish successfully
image

The animation loaded correctly, but the loaded fbx seems to be a bit off( I guess it is related to the scene_length you mentioned above)
image

client/ayon_blender/plugins/publish/extract_camera_fbx.py
moonyuet: I think we should use the global scale instead and remove the related setting inside the Publisher UI

client/ayon_blender/plugins/publish/extract_camera_fbx.py
```diff
@@ -49,8 +49,22 @@ def process(self, instance):
...
-        bpy.context.scene.unit_settings.scale_length = 0.01
+        scene = bpy.context.scene
+        scale_length = scene.unit_settings.scale_length
+        frame_start = scene.frame_start
+        frame_end = scene.frame_end
+        frame_step = scene.frame_step
+        fps = scene.render.fps
+        fps_base = scene.render.fps_base
+        scene.unit_settings.scale_length = instance.data.get(
+            "unitScale", scale_length)
```
moonyuet:
```suggestion
```

client/ayon_blender/plugins/publish/extract_camera_fbx.py
```diff
@@ -66,7 +80,12 @@ def process(self, instance):
                 bake_anim_simplify_factor=0.0
             )
 
-        bpy.context.scene.unit_settings.scale_length = scale_length
+        scene.unit_settings.scale_length = scale_length
```
moonyuet:
```suggestion
```

client/ayon_blender/plugins/publish/extract_fbx.py
```diff
@@ -55,8 +55,9 @@ def process(self, instance):
                 new_materials.append(mat)
                 new_materials_objs.append(obj)
 
-        scale_length = bpy.context.scene.unit_settings.scale_length
-        bpy.context.scene.unit_settings.scale_length = 0.01
+        scene = bpy.context.scene
+        scale_length = scene.unit_settings.scale_length
+        scene.unit_settings.scale_length = instance.data.get("unitScale")
```
moonyuet:
```suggestion
```

client/ayon_blender/plugins/publish/extract_fbx.py
```diff
@@ -68,7 +69,7 @@ def process(self, instance):
                 add_leaf_bones=False
             )
 
-        bpy.context.scene.unit_settings.scale_length = scale_length
+        scene.unit_settings.scale_length = scale_length
```
moonyuet:
```suggestion
```

client/ayon_blender/plugins/publish/extract_camera_fbx.py
```diff
@@ -49,24 +50,36 @@ def process(self, instance):
...
-
-        bpy.context.scene.unit_settings.scale_length = scale_length
+        attr_values = self.get_attr_values_from_data(instance.data)
+        scene_overrides = {
+            "frame_start": instance.data.get("frameStart"),
+            "frame_end": instance.data.get("frameEnd"),
+            "frame_step": instance.data.get("frameStep"),
+            "unit_settings.scale_length": attr_values.get("unitScale"),
+            "render.fps": instance.data.get("fps")
+        }
```
moonyuet, line 60:
```suggestion
        scene_overrides = {
            "frame_start": instance.data.get("frameStart"),
            "frame_end": instance.data.get("frameEnd"),
            "frame_step": instance.data.get("frameStep"),
            "render.fps": instance.data.get("fps")
        }
```

client/ayon_blender/plugins/publish/extract_fbx.py
```diff
@@ -55,20 +56,22 @@ def process(self, instance):
...
+        with lib.attribute_overrides(bpy.context.scene, scene_overrides):
+            with bpy.context.temp_override(**context):
+                # We export the fbx
+                bpy.ops.export_scene.fbx(
+                    filepath=filepath,
+                    use_active_collection=False,
+                    use_selection=True,
+                    mesh_smooth_type='FACE',
+                    add_leaf_bones=False
+                )
```
moonyuet, line 74:
```suggestion
        with bpy.context.temp_override(**context):
            # We export the fbx
            bpy.ops.export_scene.fbx(
                filepath=filepath,
                use_active_collection=False,
                use_selection=True,
                mesh_smooth_type='FACE',
                add_leaf_bones=False
            )
```

moonyuet: I find out that it ends up with some frame range issue when loaded back to maya. We still need to have scene_overrides for storing the frame range data.
```suggestion
        scene_overrides = {
            "frame_start": instance.data.get("frameStart"),
            "frame_end": instance.data.get("frameEnd"),
            "frame_step": instance.data.get("frameStep"),
            "render.fps": instance.data.get("fps")
        }
        # Skip None value overrides
        scene_overrides = {
            key: value for key, value in scene_overrides.items()
            if value is not None
        }
        if "render.fps" in scene_overrides:
            scene_overrides["render.fps_base"] = 1

        with lib.attribute_overrides(bpy.context.scene, scene_overrides):
            with bpy.context.temp_override(**context):
```

client/ayon_blender/plugins/publish/extract_fbx.py
```diff
@@ -91,3 +87,14 @@ def process(self, instance):
...
+
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            NumberDef("unitScale",
+                      label="Unit Scale (FBX)",
+                      default=1.0,
+                      decimals=4,
+                      tooltip="Scale of the model, valid only for FBX export.")
+        ]
```
moonyuet, line 99:
```suggestion

```
We dont need this anymore.

client/ayon_blender/plugins/publish/extract_camera_fbx.py
```diff
@@ -83,3 +94,14 @@ def process(self, instance):
...
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            NumberDef("unitScale",
+                      label="Unit Scale (FBX)",
+                      default=1.0,
+                      decimals=4,
+                      tooltip="Scale of the model, valid only for FBX export.")
+        ]
+
```
moonyuet, line 107:
```suggestion
```

moonyuet (commented):
The scale issue has been resolved but there is still frame range issue from extract fbx. The frame range issue does not happen in the extract camera(fbx)

LiborBatek:
@moonyuet speaking of animation and offset be aware of the ```Fps``` as it has huge impact on timeline aka loading animations for e.g. ```25fps``` vs ```24fps``` , just sayin...doesnt mean thats the issue atm

moonyuet:
> @moonyuet speaking of animation and offset be aware of the `Fps` as it has huge impact on timeline aka loading animations for e.g. `25fps` vs `24fps` , just sayin...doesnt mean thats the issue atm

If this is dangerous, guess we should also need to remove the fps option in the scene overrides so that users won't publish assets with weird fps type.

client/ayon_blender/plugins/publish/extract_abc_animation.py
```diff
@@ -55,13 +55,28 @@ def process(self, instance):
...
-            bpy.ops.wm.alembic_export(
-                filepath=filepath,
-                selected=True,
-                flatten=False
-            )
+        scene_overrides = {
+            "frame_start": instance.data.get("frameStart"),
+            "frame_end": instance.data.get("frameEnd"),
+            "frame_step": instance.data.get("frameStep"),
+            "render.fps": instance.data.get("fps")
```
moonyuet, line 62:
```suggestion

```

client/ayon_blender/plugins/publish/extract_camera_abc.py
```diff
@@ -45,13 +45,29 @@ def process(self, instance):
...
-                filepath=filepath,
-                selected=True,
-                flatten=True
-            )
+        scene_overrides = {
+            "frame_start": instance.data.get("frameStart"),
+            "frame_end": instance.data.get("frameEnd"),
+            "frame_step": instance.data.get("frameStep"),
+            "unit_settings.scale_length": instance.data.get("unitScale"),
+            "render.fps": instance.data.get("fps")
```
moonyuet, line 53:
```suggestion

```

client/ayon_blender/plugins/publish/extract_camera_fbx.py
```diff
@@ -49,24 +50,34 @@ def process(self, instance):
...
-                object_types={'CAMERA'},
-                bake_anim_simplify_factor=0.0
-            )
-
-        bpy.context.scene.unit_settings.scale_length = scale_length
+        scene_overrides = {
+            "frame_start": instance.data.get("frameStart"),
+            "frame_end": instance.data.get("frameEnd"),
+            "frame_step": instance.data.get("frameStep"),
+            "render.fps": instance.data.get("fps")
```
moonyuet, line 57:
```suggestion

```

client/ayon_blender/plugins/publish/extract_abc_animation.py
```diff
@@ -55,13 +55,28 @@ def process(self, instance):
...
+        }
+        # Skip None value overrides
+        scene_overrides = {
+            key: value for key, value in scene_overrides.items()
+            if value is not None
+        }
+        if "render.fps" in scene_overrides:
+            scene_overrides["render.fps_base"] = 1
```
moonyuet, line 70:
```suggestion

```

client/ayon_blender/plugins/publish/extract_abc.py
```diff
@@ -44,14 +44,29 @@ def process(self, instance):
...
-                filepath=filepath,
-                selected=True,
-                flatten=False,
-                subdiv_schema=attr_values.get("subdiv_schema", False)
-            )
+        scene_overrides = {
+            "frame_start": instance.data.get("frameStart"),
+            "frame_end": instance.data.get("frameEnd"),
+            "frame_step": instance.data.get("frameStep"),
+            "render.fps": instance.data.get("fps")
```
moonyuet, line 51:
```suggestion
```

BigRoy:
Unassigning myself since I may not be the best to continue this without clear pointers on what is "good" Blender behavior.

Tagging @LiborBatek @moonyuet for visibility - anyone who may be a better fit? **Or** should I just include Kayla's notes - do a simple test run and then assume I haven't broken other critical blender parts?

iLLiCiTiT:
Can someone take care of this PR? @BigRoy @moonyuet ?

LiborBatek (changes requested):
When testing I can see those new parameters just on ```Camera``` instance but not on ```Animation``` nor ```Pointcache``` there are missing...
Screenshot 2024-08-29 155754



Screenshot 2024-08-29 160027

LiborBatek:
@moonyuet maybe outside of scope of this PR but as we will introduce ```frame range``` on each publish instance...I would like to point out that we completely lack ```Validate Frame range``` on the ```Context``` widget as in e.g. Maya host. Shouldnt we introduce it for blender too?!

moonyuet:
> @moonyuet maybe outside of scope of this PR but as we will introduce `frame range` on each publish instance...I would like to point out that we completely lack `Validate Frame range` on the `Context` widget as in e.g. Maya host. Shouldnt we introduce it for blender too?!

yes we should include these families to the validate frame range, but with different PR

iLLiCiTiT:
Is this ready for review?

moonyuet:
> Is this ready for review?



> Is this ready for review?

yes, @LiborBatek and @BigRoy can you review?

LiborBatek:
I can see the ```frame range``` settings for each publish instance which is good but the UI lacks ```Unit Scale``` settings...am I correct when expecting it to be shown in the publisher UI?!

Screenshot 2024-09-11 144411


Screenshot 2024-09-11 144522


been tested on these publish instances:

animation
model
pointcache
camera
review

LiborBatek (approved):
Been tested and LGTM

PR #4, opened 2024-07-08
Bugfix the render instance without workfile being saved

### LINKED ISSUE #3: Creating render instance errors out when workfile not being saved

- [X] I have searched the existing issues

### Current Behavior:

image

### Expected Behavior:

It shouldn't error out when creating render instance without any workfile.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Launch Blender without workfile
2. Create Render Instance
3. It errors out the creator

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

_No response_

### Additional context:

_No response_

moonyuet:
### Changelog Description
Resolve #3 

### Testing Notes
1. Launch Blender 
2. Create Render Instance
3. It should be saving the new scene for creating render instance if there is no workfile saved.

### DIFF
--- a/client/ayon_blender/plugins/create/create_render.py
+++ b/client/ayon_blender/plugins/create/create_render.py
@@ -2,6 +2,7 @@
 import bpy
 
 from ayon_core.lib import version_up
+from ayon_core.pipeline.context_tools import version_up_current_workfile
 from ayon_blender.api import plugin
 from ayon_blender.api.render_lib import prepare_rendering
 from ayon_blender.api.workio import save_file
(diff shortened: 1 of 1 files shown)


moonyuet:
I am not sure if it is the best solution of versioning up the current workfile (so that's why it is at a draft) so any better tweaks are welcome.

LiborBatek (approved):
All seems working and correct.

I was able to create ```render``` publish instance even when unsaved ```workfile``` as it automatically saved it for me resulting into upversioned workfile.

LGTM

image

BigRoy (approved):
Looks good to me.

Would this potentially still error if the user did `new scene` and `bpy.data.is_dirty` returned False so that `version_up_current_workfile` didn't actually end up saving anything due to the check [here](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/pipeline/context_tools.py#L594-L597). I mean, it's an edge case so would ignore it for now. But we might want to have a "force update" toggle on that function OR at least force if the scene is unsaved completely.

PR #5, opened 2024-07-09
Implementation of Validate Frame Range

moonyuet:
### Changelog Description
Implementation of Validate Frame Range

### Testing Notes
1. Launch Blender
2. Create Render Instance
3. Publish

### DIFF
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -221,14 +221,16 @@ def message_window(title, message):
     _process_app_events()
 
 
-def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
-    """Get the current task frame range and handles
+def get_frame_range(task_entity=None) -> Union[Dict[str, int], None]:
+    """Get the task entity's frame range and handles
 
     Args:
-        task_entity (dict): Task Entity.
+        task_entity (Optional[dict]): Task Entity.
+            When not provided defaults to current context task.
 
     Returns:
-        dict: with frame start, frame end, handle start, handle end.
+        Union[Dict[str, int], None]: Dictionary with
+            frame start, frame end, handle start, handle end.
     """
     # Set frame start/end
     if task_entity is None:
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -6,7 +6,7 @@ from typing import Callable, Dict, Iterator, List, Optional
 import bpy
 
 import pyblish.api
-from typing import Union, Any
+from typing import Union
 
 from ayon_core.host import (
     HostBase,
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -6,7 +6,7 @@ from typing import Callable, Dict, Iterator, List, Optional
 import bpy
 
 import pyblish.api
-import ayon_api
+from typing import Union, Any
 
 from ayon_core.host import (
     HostBase,
@@ -16,14 +16,15 @@ from ayon_core.host import (
 )
 from ayon_core.pipeline import (
     schema,
-    get_current_project_name,
-    get_current_folder_path,
     register_loader_plugin_path,
     register_creator_plugin_path,
     deregister_loader_plugin_path,
     deregister_creator_plugin_path,
     AVALON_CONTAINER_ID,
     AYON_CONTAINER_ID,
+)
+from ayon_core.pipeline.context_tools import (
+    get_current_task_entity
 )
 from ayon_core.lib import (
     Logger,
@@ -220,12 +221,34 @@ def message_window(title, message):
     _process_app_events()
 
 
-def get_folder_attributes():
-    project_name = get_current_project_name()
-    folder_path = get_current_folder_path()
-    folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
+def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
+    """Get the current task frame range and handles
 
-    return folder_entity["attrib"]
+    Args:
+        task_entity (dict): Task Entity.
+
+    Returns:
+        dict: with frame start, frame end, handle start, handle end.
+    """
+    # Set frame start/end
+    if task_entity is None:
+        task_entity = get_current_task_entity(fields={"attrib"})
+    task_attributes = task_entity["attrib"]
+    frame_start = int(task_attributes["frameStart"])
+    frame_end = int(task_attributes["frameEnd"])
+    handle_start = int(task_attributes["handleStart"])
+    handle_end = int(task_attributes["handleEnd"])
+    frame_start_handle = frame_start - handle_start
+    frame_end_handle = frame_end + handle_end
+
+    return {
+        "frameStart": frame_start,
+        "frameEnd": frame_end,
+        "handleStart": handle_start,
+        "handleEnd": handle_end,
+        "frameStartHandle": frame_start_handle,
+        "frameEndHandle": frame_end_handle,
+    }
 
 
 def set_frame_range(data):
@@ -278,7 +301,7 @@ def on_new():
     set_resolution_startup = settings.get("set_resolution_startup")
     set_frames_startup = settings.get("set_frames_startup")
 
-    data = get_folder_attributes()
+    data = get_frame_range()
 
     if set_resolution_startup:
         set_resolution(data)
@@ -299,7 +322,7 @@ def on_open():
     set_resolution_startup = settings.get("set_resolution_startup")
     set_frames_startup = settings.get("set_frames_startup")
 
-    data = get_folder_attributes()
+    data = get_frame_range()
 
     if set_resolution_startup:
         set_resolution(data)
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -303,7 +303,7 @@ def on_new():
     set_resolution_startup = settings.get("set_resolution_startup")
     set_frames_startup = settings.get("set_frames_startup")
 
-    data = get_frame_range()
+    data = get_current_task_entity()
 
     if set_resolution_startup:
         set_resolution(data)
@@ -324,7 +324,7 @@ def on_open():
     set_resolution_startup = settings.get("set_resolution_startup")
     set_frames_startup = settings.get("set_frames_startup")
 
-    data = get_frame_range()
+    data = get_current_task_entity()
 
     if set_resolution_startup:
         set_resolution(data)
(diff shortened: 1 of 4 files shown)


client/ayon_blender/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,84 @@
...
+
+        frame_range = get_frame_range(instance.data["taskEntity"])
+        scene = bpy.context.scene
+        inst_frame_start = scene.frame_start
+        inst_frame_end = scene.frame_end
+
+        if inst_frame_start is None or inst_frame_end is None:
+            raise KnownPublishError(
+                "Missing frame start and frame end on "
+                "instance to to validate."
```
BigRoy, line 50:
```suggestion
                "instance to validate."
```

client/ayon_blender/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,84 @@
...
+            report = (
+                "Frame range settings are incorrect.\n\n"
+                f"{bullet_point_errors}\n\n"
+                "You can use repair action to fix it."
+            )
+            raise PublishValidationError(report, title="Frame Range incorrect")
+
+    @classmethod
+    def repair(cls, instance):
+        frame_range = get_frame_range()
```
BigRoy, line 78:
```suggestion
        frame_range = get_frame_range(instance.data["taskEntity"])
```

client/ayon_blender/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,84 @@
...
+        frame_start_handle = frame_range["frameStartHandle"]
+        frame_end_handle = frame_range["frameEndHandle"]
+        scene = bpy.context.scene
+        scene.frame_start = frame_start_handle
+        scene.frame_end = frame_end_handle
```
BigRoy, line 83: Is there a reason this behavior is different from [`set_frame_range`](https://github.com/ynput/ayon-blender/blob/3b893990d863afa5ecbf42658b3a8cc0d689b954/client/ayon_blender/api/pipeline.py#L257-L268)?

This includes the handles, the other doesn't. Why do they differ?

---

Also side note: Can we update `set_frame_range` (because currently it will fail to set a start or end frame of value 0. ;)

Looking at the logic it should be the same as:
```python
def set_frame_range(data: dict):
    """Set frame range
    
    Arguments:
        data (dict): Frame range and FPS
            to set the current scene to.
    """
    if not data:
        return

    scene = bpy.context.scene
    if "frameStart" in data:
        scene.frame_start = data["frameStart"]
    if "frameEnd" in data:
        scene.frame_end = data["frameEnd"]
    if "fps" in data:
        fps: float = data["fps"]
        scene.render.fps = round(fps)
        scene.render.fps_base = round(fps) / fps
```
Because the current code just sets what the scene is already set to if no data is provided (so we can just skip doing anything if not provided).

client/ayon_blender/api/pipeline.py
```diff
@@ -220,12 +221,34 @@ def message_window(title, message):
...
-    folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
+def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
+    """Get the current task frame range and handles
+
+    Args:
+        task_entity (dict): Task Entity.
 
-    return folder_entity["attrib"]
+    Returns:
+        dict: with frame start, frame end, handle start, handle end.
```
BigRoy, line 231: This return type should actually match the return type above.

```suggestion
        Union[Dict[str, Any], None]: Dictionary with 
            frame start, frame end, handle start, handle end.
```

Yet, it also seems that all the values are `int` so it should be:
```suggestion
        Union[Dict[str, int], None]: Dictionary with 
            frame start, frame end, handle start, handle end.
```

client/ayon_blender/api/pipeline.py
```diff
@@ -220,12 +221,34 @@ def message_window(title, message):
...
 
-def get_folder_attributes():
-    project_name = get_current_project_name()
-    folder_path = get_current_folder_path()
-    folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
+def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
+    """Get the current task frame range and handles
+
+    Args:
+        task_entity (dict): Task Entity.
```
BigRoy, line 228:
```suggestion
        task_entity (Optional[dict]): Task Entity.
            When not provided defaults to current context task.
```

client/ayon_blender/api/pipeline.py
```diff
@@ -220,12 +221,34 @@ def message_window(title, message):
     _process_app_events()
 
 
-def get_folder_attributes():
-    project_name = get_current_project_name()
-    folder_path = get_current_folder_path()
-    folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
+def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
+    """Get the current task frame range and handles
```
BigRoy, line 225:
```suggestion
    """Get the task entity's frame range and handles
```

client/ayon_blender/api/pipeline.py
```diff
@@ -220,12 +221,34 @@ def message_window(title, message):
     _process_app_events()
 
 
-def get_folder_attributes():
-    project_name = get_current_project_name()
-    folder_path = get_current_folder_path()
-    folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
+def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
```
BigRoy, line 224:
```suggestion
def get_frame_range(task_entity=None) -> Union[Dict[str, int], None]:
```

LiborBatek (changes requested):
Doesnt work for me...it doesnt Validate the frame range correctly if it differs from Ayon DB entry...
Screenshot 2024-07-10 093440

It also seems that ```Set Frame Range``` and ```Set Resolution``` actions suffer from malfunction too

When trying to trigger ```Set Frame Range``` Im getting this:
Screenshot 2024-07-10 093500

Im also enclosing full JSON report after Publishing of the ```Render```


[publish-report-240710-09-38.json.txt](https://github.com/user-attachments/files/16157162/publish-report-240710-09-38.json.txt)

LiborBatek (changes requested):
Still behaves the same....Im free to publish even when ```frame range``` being ```incorrect```...

image

client/ayon_blender/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,78 @@
...
+            report = (
+                "Frame range settings are incorrect.\n\n"
+                f"{bullet_point_errors}\n\n"
+                "You can use repair action to fix it."
+            )
+            raise PublishValidationError(report, title="Frame Range incorrect")
+
+    @classmethod
+    def repair(cls, instance):
+        set_frame_range(instance.data["taskEntity"])
```
BigRoy, line 78: The only issue I think now is still that it appears to validate against frame handle start and frame handle end whereas the repair will set the frame range to the frame start without handles and the frame end without handles, right?

client/ayon_blender/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,78 @@
...
+        errors = []
+        if frame_start_handle != inst_frame_start:
+            errors.append(
+                f"Start frame ({inst_frame_start}) on instance does not match " # noqa
+                f"with the start frame ({frame_start_handle}) set on the folder attributes. ")    # noqa
+        if frame_end_handle != inst_frame_end:
+            errors.append(
+                f"End frame ({inst_frame_end}) on instance does not match "
+                f"with the end frame ({frame_end_handle}) "
+                "from the folder attributes. ")
```
BigRoy, line 63: I wonder if this is more correct when referring to task attributes instead of folder attributes.
```suggestion
            errors.append(
                f"Start frame ({inst_frame_start}) on instance does not match " # noqa
                f"with the start frame ({frame_start_handle}) set on the task attributes. ")    # noqa
        if frame_end_handle != inst_frame_end:
            errors.append(
                f"End frame ({inst_frame_end}) on instance does not match "
                f"with the end frame ({frame_end_handle}) "
                "from the task attributes. ")
```

client/ayon_blender/plugins/publish/validate_frame_range.py
BigRoy: Did you push this yet?

moonyuet: Yes.

client/ayon_blender/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,78 @@
...
+        errors = []
+        if frame_start_handle != inst_frame_start:
+            errors.append(
+                f"Start frame ({inst_frame_start}) on instance does not match " # noqa
+                f"with the start frame ({frame_start_handle}) set on the task attributes. ")    # noqa
+        if frame_end_handle != inst_frame_end:
+            errors.append(
+                f"End frame ({inst_frame_end}) on instance does not match "
+                f"with the end frame ({frame_end_handle}) "
+                "from the task attributes. ")
```
BigRoy, line 63:
```suggestion
        frame_start = frame_range["frameStart"]
        frame_end = frame_range["frameEnd"]
        errors = []
        if frame_start != inst_frame_start:
            errors.append(
                f"Start frame ({inst_frame_start}) on instance does not match " # noqa
                f"with the start frame ({frame_start}) set on the task attributes. ")    # noqa
        if frame_end != inst_frame_end:
            errors.append(
                f"End frame ({inst_frame_end}) on instance does not match "
                f"with the end frame ({frame_end}) "
                "from the task attributes. ")
```

BigRoy (approved):
Looks good to me by code review. @LiborBatek up for a test drive?

LiborBatek (changes requested):
The Validation works well when incorrect frame range set but the ```Repair``` action doesnt work for me still.

image


It also doesnt show any error msg...and just says ```Action Finnished```. 

When manually set correct ```Frame range``` then publishing works fine and validation not  fail.

moonyuet:
> The Validation works well when incorrect frame range set but the `Repair` action doesnt work for me still.
> 
> image
> 
> It also doesnt show any error msg...and just says `Action Finnished`.
> 
> When manually set correct `Frame range` then publishing works fine and validation not fail.

it should be working now. the repair action needs the `bpy.context.scene.frame_start` and ``bpy.context.scene.frame_end` to repair the frame range

LiborBatek (approved):
Now  ```Validate frame range``` and ```Repair``` action works as supposed to!

LGTM

image

PR #7, opened 2024-07-12
Allow existing QtWidgets.QApplication instance to exist

BigRoy:
### Issue

Another Blender addon or script may, before AYON's initialization scripts run have already done other things like initializing a `QApplication` instance. We should allow that and not error on such a conflict with another addon.

### Note

It's now not inheriting from QApplication anymore because we weren't using the instance itself really anywhere else - it was all classmethod access anyway. 

### Additional context

This may fix [the issue reported by Luciano Munoz on Ynput Discord](https://discord.com/channels/517362899170230292/1260337785403867218/1261001453581897758)

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -204,7 +204,6 @@ def _process_app_events() -> Optional[float]:
 class LaunchQtApp(bpy.types.Operator):
     """A Base class for operators to launch a Qt app."""
 
-    _app: QtWidgets.QApplication
     _window = Union[QtWidgets.QDialog, ModuleType]
     _tool_name: str = None
     _init_args: Optional[List] = list()
@@ -215,8 +214,7 @@ class LaunchQtApp(bpy.types.Operator):
         if self.bl_idname is None:
             raise NotImplementedError("Attribute `bl_idname` must be set!")
         print(f"Initialising {self.bl_idname}...")
-        self._app = BlenderApplication.get_app()
-        GlobalClass.app = self._app
+        GlobalClass.app = BlenderApplication.get_app()
 
         if not bpy.app.timers.is_registered(_process_app_events):
             bpy.app.timers.register(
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Tested in Blender with loader, publisher, workfile etc.
Still work as expected.

BigRoy:
This PR [fixes the issue Luciano Munoz faced](https://discord.com/channels/517362899170230292/1260337785403867218/1261381239684726995)

Look at this Blender beauty:
blender_beauty_by_luciano_munoz

client/ayon_blender/api/ops.py
```diff
@@ -41,23 +41,31 @@ def wrapper(*args, **kwargs):
...
+            # Ensure it is configured to our needs
+            cls._prepare_qapplication(application)
+            cls._instance = application
+
         return cls._instance
 
+    @classmethod
+    def _prepare_qapplication(cls, application: QtWidgets.QApplication):
+        application.setQuitOnLastWindowClosed(False)
+        application.setStyleSheet(style.load_stylesheet())
```
iLLiCiTiT, line 66: If there is existing application, would this possible break their styles?

BigRoy: Yes - we're basically overriding the `QApplication` style, so any other widgets by any other Qt tool would also inherit those styles from the QApplication. Yes.

iLLiCiTiT: Because I do expect that the error happened because other addon created QApplication, so we might break styles of other addon (or other addon can break our styles).

I'm curious if we need to set the styles for whole application?

BigRoy: Yup - I'm also curious as to why that logic is there. I believe all widgets we're opening get their stylesheets applied in a dedicated manner anyway so we should be able to rely on that.

iLLiCiTiT: Maybe there are popup dialogs that don't set it?

PR #12, opened 2024-07-18
Do not error on unsaved workfile

BigRoy:
This is an extension of another PR in ayon-core: https://github.com/ynput/ayon-core/pull/786

Together they generate a better validation report for when the publish started from an unsaved workfile.

image

### DIFF
--- a/client/ayon_blender/plugins/publish/collect_workfile.py
+++ b/client/ayon_blender/plugins/publish/collect_workfile.py
@@ -16,7 +16,7 @@ class CollectWorkfile(plugin.BlenderInstancePlugin):
         """Process collector."""
 
         context = instance.context
-        filepath = Path(context.data["currentFile"])
+        filepath = Path(context.data.get("currentFile") or "")
         ext = filepath.suffix
 
         instance.data.update(
--- a/client/ayon_blender/plugins/publish/collect_workfile.py
+++ b/client/ayon_blender/plugins/publish/collect_workfile.py
@@ -16,7 +16,15 @@ class CollectWorkfile(plugin.BlenderInstancePlugin):
         """Process collector."""
 
         context = instance.context
-        filepath = Path(context.data.get("currentFile") or "")
+        filepath = context.data.get("currentFile")
+        if not filepath:
+            self.log.warning("Deactivating workfile instance because no "
+                             "current filepath is found. Please save your "
+                             "workfile.")
+            instance.data["publish"] = False
+            return
+
+        filepath = Path(filepath)
         ext = filepath.suffix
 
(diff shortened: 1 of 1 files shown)


client/ayon_blender/plugins/publish/collect_workfile.py
```diff
@@ -16,7 +16,7 @@ def process(self, instance):
         """Process collector."""
 
         context = instance.context
-        filepath = Path(context.data["currentFile"])
+        filepath = Path(context.data.get("currentFile") or "")
         ext = filepath.suffix
 
         instance.data.update(
```
moonyuet, line 22: if the filepath is empty string, `ext =filepath.suffix` most likely errors out.

BigRoy: I did a test run and it seemed fine. ✅ 

It would maybe be better if current file is none that it'd not end up doing the rest and return - but I'm worried that other plug-ins may be dependent on those attributes and hence we'll run into other errors.

iLLiCiTiT: Shouldn't be the instance deactivated if filepath is not available?

iLLiCiTiT: BTW the paht is filled in `"setMembers"` but I couldn't find usage of the key anywhere.

BigRoy: This is actually used automatically due to the deadline logic using the `AbstractSubmitDeadline`?

https://github.com/ynput/ayon-deadline/blob/c386aeb2e83f8e484c19d373d57e6ff561ff14b7/client/ayon_deadline/plugins/publish/blender/submit_blender_deadline.py#L29-L31

BigRoy: > Shouldn't be the instance deactivated if filepath is not available?

Valid point - we could just set the instance data active to False. That's quite clever actually.

iLLiCiTiT: It was just first idea, but I don't know hat are consequences? If this add workfile to render instances then it might not be good idea, and maybe let it crash is better actually.

BigRoy: > If this add workfile to render instances then it might not be good idea, and maybe let it crash is better actually.

A validator still disallows the actual publish - this is just to avoid errors on Collecting. Made the changes, seemed fine for me.

moonyuet (approved):
Looks good for me.
image

Issue #13, opened 2024-07-22
Add possibility to pack texture or not before publish.


jeremy-bepoix:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Currently, if you publish an asset, the textures are packed into the blend file. If you are not using site-sync or remote is non sense because all textures can be found in absolute, plus it makes your blender file considerably heavier.
[See : ](https://github.com/ynput/OpenPype/pull/2085)

Another thing, if you use image sequences in texture nodes and they don't start with 00, this breaks the texture package and won't publish too.

### How would you imagine the implementation of the feature?

Add toogle on frontend : Pack used images in the blend files.
mokup : 
![image](https://github.com/user-attachments/assets/fc3642ba-4705-46e1-b26c-ca4a9e3d01ab)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Additional context:

Also see discussion here: https://discord.com/channels/517362899170230292/1248298172816822424/1263781294421315655

BigRoy:
> Another thing, if you use image sequences in texture nodes and they don't start with 00, this breaks the texture package and won't publish too.

Interesting bug! :) Do you have more background details to this? Is this a known blender bug or a known limitation? Why does it fail?

---

Do you think it's worth having the settings define the default but potentially allowing to have an artist-facing toggle in the publisher UI so that if just one particular publish needed it, they can enable it. Or is defining the value in studio settings sufficient for production usage?

jeremy-bepoix:
> Like the code they link @jeremy-bepoix - does it log a warning or error? Or does it 'pack' but just fail when loading?
Yep, is known limitation This produces an error if the start frame does not begin with 00
see : [bpy.types.Image.pack()](https://docs.blender.org/api/current/bpy.types.Image.html#bpy.types.Image.pack)

It's not easy in the sense that if you start to put your finger in the resource pack, it's going to get a bit complicated because you can pack each blender entitites. Pack Linked Libraries, pack Data, Pack Resources ect...
Then you can also un-pack..

The pack should be seen as a kind of archive. This could be useful for sending to a client, or a freelancer who doesn't have all the studio's resources.

I think it's best not to pack anything in the first place and maybe force absolute path for all files instead.
[...]

PR #14, opened 2024-07-24
Validator to check on external data path being absolute

moonyuet:
## Changelog Description
This PR is to add the validator to check the external data path being absolute to ensure that the path would not break the links during publishing.

## Additional info
Build the blender addon with this branch as  the validator has been added into the ayon setting.
image

## Testing notes:

1. Launch Blender
2. Publish
3. If the external data paths are relative, the validator would error out
4. Use repair action
5. It will turn the external data paths to be absolute

### DIFF
--- a/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
+++ b/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
@@ -1,5 +1,4 @@
 import bpy
-import inspect
 import pyblish.api
 from ayon_core.pipeline import (
     OptionalPyblishPluginMixin
--- a/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
+++ b/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
@@ -69,7 +69,6 @@ class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
 
             Using the Repair action will turn all datablock filepaths in your scene into
             absolute filepaths.
-
         """)
 
     @classmethod
--- a/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
+++ b/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
@@ -61,7 +61,7 @@ class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
 
     @classmethod
     def get_description(cls):
-        return inspect.cleandoc(f"""
+        return inspect.cleandoc("""
             ### Data block filepaths are invalid
             Data block filepaths must be absolute paths to avoid issues during relocation
             of the published workfile into the publish folder.
--- a/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
+++ b/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
@@ -1,4 +1,5 @@
 import bpy
+import inspect
 import pyblish.api
 from ayon_core.pipeline import (
     OptionalPyblishPluginMixin
@@ -15,12 +16,13 @@ class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
     """Validates Absolute Data Block Paths
 
     This validator checks if all external data paths are absolute
-    to ensure the links would not be broke when publishing
+    to ensure the links would not be broken when publishing
     """
 
     label = "Validate Absolute Data Block Paths"
     order = ValidateContentsOrder
     hosts = ["blender"]
+    families = ["workfile"]
     optional = True
     actions = [RepairContextAction]
 
@@ -30,18 +32,19 @@ class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
         object_type = type(bpy.data.objects)
         for attr in dir(bpy.data):
             collections = getattr(bpy.data, attr)
-            if isinstance(collections, object_type):
-                for data_block in collections:
-                    if not hasattr(data_block, "filepath"):
-                        continue
-                    if not data_block.filepath:
-                        continue
-                    if data_block.filepath == bpy.path.abspath(data_block.filepath):
-                        continue
+            if not isinstance(collections, object_type):
+                continue
+            for data_block in collections:
+                if not hasattr(data_block, "filepath"):
+                    continue
+                if not data_block.filepath:
+                    continue
+                if data_block.filepath == bpy.path.abspath(data_block.filepath):
+                    continue
 
-                    cls.log.error(f"Data block filepath {data_block.filepath} "
-                                    "is not absolute path")
-                    invalid.append(data_block.filepath)
+                cls.log.error(f"Data block filepath {data_block.filepath} "
+                                "is not absolute path")
+                invalid.append(data_block.filepath)
         return invalid
 
     def process(self, context):
@@ -53,12 +56,22 @@ class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
             bullet_point_errors = "\n".join(
                 "- {}".format(error) for error in invalid
             )
-            report = (
-                "Data block filepaths are not absolute.\n\n"
-                f"{bullet_point_errors}\n\n"
-                "You can use repair action to fix it."
+            raise PublishValidationError(
+                "Invalid Data block filepaths",
+                title="Relative Data block filepaths",
+                description=self.get_description(bullet_point_errors)
             )
-            raise PublishValidationError(report, title="Relative Data block filepaths")
+
+    @classmethod
+    def get_description(cls, bullet_point_errors):
+        return inspect.cleandoc(f"""
+            ### Data block filepaths are invalid
+
+            Data block filepaths are not in absolute paths:
+            #### {bullet_point_errors}\n
+            You can use repair action to make the data block filepaths absolute.
+
+        """)
 
     @classmethod
     def repair(cls, context):
--- a/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
+++ b/client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
@@ -4,58 +4,62 @@ from ayon_core.pipeline import (
     OptionalPyblishPluginMixin
 )
 from ayon_core.pipeline.publish import (
-    RepairAction,
+    RepairContextAction,
     ValidateContentsOrder,
     PublishValidationError
 )
 
 
-
-class ValidateAbsoluteExternalDataPaths(pyblish.api.InstancePlugin,
-                                        OptionalPyblishPluginMixin):
-    """Validates Absolute External Data Paths
+class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
+                                     OptionalPyblishPluginMixin):
+    """Validates Absolute Data Block Paths
 
     This validator checks if all external data paths are absolute
     to ensure the links would not be broke when publishing
     """
 
-    label = "Validate Absolute External Data Paths"
+    label = "Validate Absolute Data Block Paths"
     order = ValidateContentsOrder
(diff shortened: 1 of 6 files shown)


client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,61 @@
...
+        invalid = []
+        object_type = type(bpy.data.objects)
+        for attr in dir(bpy.data):
+            collections = getattr(bpy.data, attr)
+            if isinstance(collections, object_type):
+                for data_block in collections:
+                    if hasattr(data_block, "filepath") and data_block.filepath:
+                        if data_block.filepath != bpy.path.abspath(data_block.filepath):
+                            cls.log.error(f"Absolute external data path {data_block.filepath} "
+                                          "is not absolute path")
```
BigRoy, line 39:
```suggestion
                            cls.log.error(f"External data path {data_block.filepath} "
                                          "is not absolute path")
```

Also, I wonder if referring to these as "Data block filepath" makes more sense since that's apparently what blender calls them instead of "External data path"?

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,61 @@
...
+import pyblish.api
+from ayon_core.pipeline import (
+    OptionalPyblishPluginMixin
+)
+from ayon_core.pipeline.publish import (
+    RepairAction,
+    ValidateContentsOrder,
+    PublishValidationError
+)
+
```
BigRoy, line 11:
```suggestion
```

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,61 @@
...
+    This validator checks if all external data paths are absolute
+    to ensure the links would not be broke when publishing
+    """
+
+    label = "Validate Absolute External Data Paths"
+    order = ValidateContentsOrder
+    hosts = ["blender"]
+    optional = True
+    actions = [RepairAction]
```
BigRoy, line 26: This runs over all instances, but seems to do a scene-wide check not related to a particular instance? So it should instead be a ContextPlugin to avoid running the same check multiple times.

Plus, making this optional would mean you'd only need to disable once then instead of one each instance (where it would be validating the same thing).

So, either:
- `InstancePlugin` -> `ContextPlugin` - you may need a `RepairContextAction`
- Make this instance specific and only validate data blocks relevant to the instance.

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,61 @@
...
+    @classmethod
+    def get_invalid(cls, instance):
+        invalid = []
+        object_type = type(bpy.data.objects)
+        for attr in dir(bpy.data):
+            collections = getattr(bpy.data, attr)
+            if isinstance(collections, object_type):
+                for data_block in collections:
+                    if hasattr(data_block, "filepath") and data_block.filepath:
+                        if data_block.filepath != bpy.path.abspath(data_block.filepath):
```
BigRoy, line 37: This would be much less indented (and more readable?) if not using the deep nesting but instead `continue` if something is not what we want.

For example:
```python
        object_type = type(bpy.data.objects)
        for attr in dir(bpy.data):
            collections = getattr(bpy.data, attr)
            if not isinstance(collections, object_type):
                continue
            for data_block in collections:
                if not hasattr(data_block, "filepath") or not data_block.filepath:
                    continue

                if data_block.filepath == bpy.path.abspath(data_block.filepath):
                    continue
                    
                # etc.
```

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,65 @@
...
+    PublishValidationError
+)
+
+
+class ValidateAbsoluteDataBlockPaths(pyblish.api.ContextPlugin,
+                                     OptionalPyblishPluginMixin):
+    """Validates Absolute Data Block Paths
+
+    This validator checks if all external data paths are absolute
+    to ensure the links would not be broke when publishing
```
BigRoy, line 18:
```suggestion
    to ensure the links would not be broken when publishing
```

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,65 @@
...
+        if invalid:
+            bullet_point_errors = "\n".join(
+                "- {}".format(error) for error in invalid
+            )
+            report = (
+                "Data block filepaths are not absolute.\n\n"
+                f"{bullet_point_errors}\n\n"
+                "You can use repair action to fix it."
+            )
+            raise PublishValidationError(report, title="Relative Data block filepaths")
```
BigRoy, line 61: Preferably the report also include a "Description" describing WHY this needs to be validated and HOW the repair action fixes it, etc.

The docstring on the plug-in to me says it will be broken on publishing? Why? What specifically breaks it? Is it that we're PACKING the datablocks for images (and that is the root of the issue?) or is it ONLY relevant for workfile publishes when we're actually copying the workfile - If so, should it only run over `workfile` family? Etc.

It lacks some details as to why this validation is relevant to perform. A description could help there.

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,65 @@
...
+                    if not hasattr(data_block, "filepath"):
+                        continue
+                    if not data_block.filepath:
+                        continue
+                    if data_block.filepath == bpy.path.abspath(data_block.filepath):
+                        continue
+
+                    cls.log.error(f"Data block filepath {data_block.filepath} "
+                                    "is not absolute path")
+                    invalid.append(data_block.filepath)
```
BigRoy, line 44:
```suggestion
            if not isinstance(collections, object_type):
                continue
            for data_block in collections:
                if not hasattr(data_block, "filepath"):
                    continue
                if not data_block.filepath:
                    continue
                if data_block.filepath == bpy.path.abspath(data_block.filepath):
                    continue

                cls.log.error(f"Data block filepath {data_block.filepath} "
                                "is not absolute path")
                invalid.append(data_block.filepath)
```

BigRoy (approved):
Code make sense (didn't do a test though!)

However, I did have some mostly cosmetic notes - like the lack of a decent description for the validator.

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,78 @@
...
+            for data_block in collections:
+                if not hasattr(data_block, "filepath"):
+                    continue
+                if not data_block.filepath:
+                    continue
+                if data_block.filepath == bpy.path.abspath(data_block.filepath):
+                    continue
+
+                cls.log.error(f"Data block filepath {data_block.filepath} "
+                                "is not absolute path")
```
BigRoy, line 46:
```suggestion
                              "is not absolute path")
```

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,78 @@
...
+                title="Relative Data block filepaths",
+                description=self.get_description(bullet_point_errors)
+            )
+
+    @classmethod
+    def get_description(cls, bullet_point_errors):
+        return inspect.cleandoc(f"""
+            ### Data block filepaths are invalid
+
+            Data block filepaths are not in absolute paths:
```
BigRoy, line 70:
```suggestion
            Data block filepaths are not absolute paths:
```

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,78 @@
...
+            #### {bullet_point_errors}\n
+            You can use repair action to make the data block filepaths absolute.
```
BigRoy, line 72: Pretty sure you don't want to put these entries behind `####` making them a header?

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,78 @@
...
+
+        """)
```
BigRoy, line 74: As far as I know the idea is usually that the description is "static" content as an explanation of the validator and what is amiss wheras the actual logs from the plug-in report the actual state of your scene, e.g. the invalid filepaths, etc.

That's how I always understood it at least... @mkolar @antirotor might be able to confirm/deny?
```suggestion
            ### Data block filepaths are invalid

            Data block filepaths must be absolute paths to avoid issues during relocation 
            of the published workfile into the publish folder.
            
            #### How to repair?
            
            Using the Repair action will turn all datablock filepaths in your scene into 
            absolute filepaths.
        """)
```

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,77 @@
...
+        return ("""
+            ### Data block filepaths are invalid
+            Data block filepaths must be absolute paths to avoid issues during relocation
+            of the published workfile into the publish folder.
+
+            #### How to repair?
+
+            Using the Repair action will turn all datablock filepaths in your scene into
+            absolute filepaths.
+
```
BigRoy, line 72:
```suggestion
```

BigRoy (approved):
Code looks good to me.

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
```diff
@@ -0,0 +1,76 @@
...
+        return ("""
+            ### Data block filepaths are invalid
+            Data block filepaths must be absolute paths to avoid issues during relocation
+            of the published workfile into the publish folder.
+
+            #### How to repair?
+
+            Using the Repair action will turn all datablock filepaths in your scene into
+            absolute filepaths.
+        """)
```
BigRoy, line 72: Does this work nicely out of the box? I'd have expected this would need a `inspect.cleandocs()` call as well to avoid the indentations of the string being problematic in the markdown formatting.

LiborBatek (approved):
This PR do what it should...all working fine.

When relative paths been in use it pops up message and ```repair``` action fixes it.

Screenshot 2024-07-30 101505

Screenshot 2024-07-30 101542




P.S. small note tho, the relative vs absolute paths usage being sporadical as when publishing ```rig``` for example it packs all textures into the asset file (aka scene file) so its not referencing anything externally. This is something we should investigate more speaking of other product types like ```model``` etc.

However this feature has usage when ```linked``` data aka scene files been in use as its crucial to use such linking in absolute path manner for sake of farm rendering...

overall LGTM

LiborBatek:
@moonyuet btw does this validator work for any linked data sets or just those present in the publish instance(s) ?! as it might be useful too for whole scene file

I guess I will test it more...when using linked scenes in lighting scene for example...just to see for myself :)

BigRoy:
> P.S. small note tho, the relative vs absolute paths usage being sporadical as when publishing rig for example it packs all textures into the asset file (aka scene file) so its not referencing anything externally. This is something we should investigate more speaking of other product types like model etc.

Be aware there were some recent comments about the "packing" of external data and issues blender may have with that. See: https://github.com/ynput/ayon-blender/issues/13

client/ayon_blender/plugins/publish/validate_absolute_external_data_paths.py
moonyuet: I dont think it has been fixed at all without insepct.cleandocs()
image
With the inspect.cleandocs
image

BigRoy (approved):
Go! 🚀

PR #15, opened 2024-07-25
add set unit scale options into AYON menu

moonyuet:
## Changelog Description
This PR is to add unit scale options into AYON menu

## Additional info
Make sure to turn on `ayon+settings://blender/unit_scale_settings/enabled` 
image

We need to consider whether `ayon+settings://blender/unit_scale_settings/apply_on_opening` is required for some enhancement, as the window of log message does not show on top.

image

## Testing notes:

1. Launch Blender
2. AYON -> Set Unit Scale
image

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -430,7 +430,11 @@ class TOPBAR_MT_avalon(bpy.types.Menu):
         layout.separator()
         layout.operator(SetFrameRange.bl_idname, text="Set Frame Range")
         layout.operator(SetResolution.bl_idname, text="Set Resolution")
-        layout.operator(SetUnitScale.bl_idname, text="Set Unit Scale")
+        project = os.environ.get("AYON_PROJECT_NAME")
+        settings = get_project_settings(project).get("blender")
+        unit_scale_settings = settings.get("unit_scale_settings")
+        if unit_scale_settings.get("enabled"):
+            layout.operator(SetUnitScale.bl_idname, text="Set Unit Scale")
         layout.separator()
         layout.operator(LaunchWorkFiles.bl_idname, text="Work Files...")
 
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -387,7 +387,7 @@ class SetUnitScale(bpy.types.Operator):
         project = os.environ.get("AYON_PROJECT_NAME")
         settings = get_project_settings(project).get("blender")
         unit_scale_settings = settings.get("unit_scale_settings")
-        pipeline.set_unit_scale_for_setting(
+        pipeline.set_unit_scale_from_settings(
             unit_scale_settings=unit_scale_settings)
         return {"FINISHED"}
 
@@ -431,11 +431,7 @@ class TOPBAR_MT_avalon(bpy.types.Menu):
         layout.separator()
         layout.operator(SetFrameRange.bl_idname, text="Set Frame Range")
         layout.operator(SetResolution.bl_idname, text="Set Resolution")
-        project = os.environ.get("AYON_PROJECT_NAME")
-        settings = get_project_settings(project).get("blender")
-        unit_scale_settings = settings.get("unit_scale_settings")
-        if unit_scale_settings.get("enabled"):
-            layout.operator(SetUnitScale.bl_idname, text="Set Unit Scale")
+        layout.operator(SetUnitScale.bl_idname, text="Set Unit Scale")
         layout.separator()
         layout.operator(LaunchWorkFiles.bl_idname, text="Work Files...")
 
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -187,9 +187,6 @@ def _process_app_events() -> Optional[float]:
 
         # Refresh Manager
         if GlobalClass.app:
-           if not hasattr(GlobalClass.app, "get_window"):
-                GlobalClass.app.processEvents()
-                return TIMER_INTERVAL
            manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
            if manager:
                manager.refresh()
@@ -390,7 +387,8 @@ class SetUnitScale(bpy.types.Operator):
         project = os.environ.get("AYON_PROJECT_NAME")
         settings = get_project_settings(project).get("blender")
         unit_scale_settings = settings.get("unit_scale_settings")
-        pipeline.set_unit_scale(unit_scale_settings)
+        pipeline.set_unit_scale_for_setting(
+            unit_scale_settings=unit_scale_settings)
         return {"FINISHED"}
 
 
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -187,9 +187,12 @@ def _process_app_events() -> Optional[float]:
 
         # Refresh Manager
         if GlobalClass.app:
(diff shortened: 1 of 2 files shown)


m-u-r-p-h-y:
Is adding this option to the menu solving the issue of the wrong scale when importing alembics from Maya?

moonyuet:
> Is adding this option to the menu solving the issue of the wrong scale when importing alembics from Maya?

Not entirely, users need to set the unit scale first and then import alembic to make sure it is correct, but this would also mean the fbx import can be possibly wrong.

We can add options to multiplying 0.01 of the scale if it is from maya. But along with this unit scale setting, it can be somehow messed up the scale (depending on the unit scale being enabled or not).
I can set the condition if the unit scale being enabled and the unit scale is 0.01 i won't use the alembic import option.

client/ayon_blender/api/ops.py
```diff
@@ -186,9 +187,12 @@ def _process_app_events() -> Optional[float]:
...
         if GlobalClass.app:
-            manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
-            if manager:
-                manager.refresh()
+           if not hasattr(GlobalClass.app, "get_window"):
+                GlobalClass.app.processEvents()
+                return TIMER_INTERVAL
+           manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
+           if manager:
+               manager.refresh()
```
BigRoy, line 195: Why is this changing in this PR? Was there a bug?

client/ayon_blender/api/pipeline.py
```diff
@@ -296,6 +296,15 @@ def set_resolution(data):
     scene.render.resolution_y = resolution_y
 
 
+def set_unit_scale(unit_scale_settings=None):
+    if unit_scale_settings is None:
+        return
+    unit_scale_enabled = unit_scale_settings.get("enabled")
+    if unit_scale_enabled:
+        unit_scale = unit_scale_settings.get("base_file_unit_scale")
+        bpy.context.scene.unit_settings.scale_length = unit_scale
```
BigRoy, line 305: I think this being a function is redundant. It's odd with this argument - I'd expect it to be the scale you want to set, regardless of whether a setting is enabled or not.

MilaKudr (approved):

client/ayon_blender/api/ops.py
moonyuet: yes, the pop-up dialog showed correctly when I opened last workfile during launch but it doesn't show up on top correctly when I use workfile tool to open the scene. Maybe we can fix it with different PRs.

client/ayon_blender/api/ops.py
```diff
@@ -418,6 +431,11 @@ def draw(self, context):
         layout.separator()
         layout.operator(SetFrameRange.bl_idname, text="Set Frame Range")
         layout.operator(SetResolution.bl_idname, text="Set Resolution")
+        project = os.environ.get("AYON_PROJECT_NAME")
+        settings = get_project_settings(project).get("blender")
+        unit_scale_settings = settings.get("unit_scale_settings")
+        if unit_scale_settings.get("enabled"):
+            layout.operator(SetUnitScale.bl_idname, text="Set Unit Scale")
```
BigRoy, line 438: This seems odd actually - shouldn't the menu entry itself ALWAYS be present - even when not enabled to "automatically set"?
```suggestion
        layout.operator(SetUnitScale.bl_idname, text="Set Unit Scale")
```

client/ayon_blender/api/pipeline.py
```diff
@@ -296,6 +296,15 @@ def set_resolution(data):
     scene.render.resolution_y = resolution_y
 
 
+def set_unit_scale_for_setting(unit_scale_settings=None):
```
BigRoy, line 299: It's not setting it "for" a "setting" but from the settings?
```suggestion
def set_unit_scale_from_settings(unit_scale_settings=None):
```

client/ayon_blender/api/ops.py
```diff
@@ -378,6 +379,18 @@ def execute(self, context):
         pipeline.set_resolution(data)
         return {"FINISHED"}
 
+class SetUnitScale(bpy.types.Operator):
+    bl_idname = "wm.ayon_set_unit_scale"
+    bl_label = "Set Unit Scale"
+
+    def execute(self, context):
+        project = os.environ.get("AYON_PROJECT_NAME")
```
BigRoy, line 387: Getting the current project name should not rely solely on `os.environ`. Don't we have dedicated functions to "get the current project name"? We should rely on those.

I believe there's even a `get_current_project_settings` function somewhere to import as well.

BigRoy (changes requested):

client/ayon_blender/api/ops.py
```diff
@@ -186,9 +187,9 @@ def _process_app_events() -> Optional[float]:
 
         # Refresh Manager
         if GlobalClass.app:
-            manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
-            if manager:
-                manager.refresh()
+           manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
+           if manager:
+               manager.refresh()
```
BigRoy, line 192: Is this indentation wrong?
```suggestion
            manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
            if manager:
                manager.refresh()
```

client/ayon_blender/api/ops.py
```diff
@@ -378,6 +379,18 @@ def execute(self, context):
         pipeline.set_resolution(data)
         return {"FINISHED"}
```
BigRoy, line 381: Please check your linter :)
```suggestion


```

client/ayon_blender/api/ops.py
BigRoy: This is marked resolved, but code still seems old?

client/ayon_blender/api/pipeline.py
```diff
@@ -296,8 +297,17 @@ def set_resolution(data):
     scene.render.resolution_y = resolution_y
 
 
+def set_unit_scale_from_settings(unit_scale_settings=None):
+    if unit_scale_settings is None:
+        return
+    unit_scale_enabled = unit_scale_settings.get("enabled")
+    if unit_scale_enabled:
+        unit_scale = unit_scale_settings.get("base_file_unit_scale")
+        bpy.context.scene.unit_settings.scale_length = unit_scale
```
BigRoy, line 306: If this key did not exist, we're setting `None` to the blender value which I'm _pretty sure_ would not work? :)

```suggestion
        unit_scale = unit_scale_settings["base_file_unit_scale"]
        bpy.context.scene.unit_settings.scale_length = unit_scale
```

Should we just require the key to exist?

LiborBatek (commented):
It works as expected, been tested with different options and all working fine.

There is small UI glitch on ayon server settings in the ```value box``` when trying to enter decimal values its really pain to do so (entering value like ```0.01``` ) I guess we have encountered this before and its tied to the box format (type of data) need to be changed I guess...
Screenshot 2024-07-29 173718

BigRoy:
> There is small UI glitch on ayon server settings in the `value box` when trying to enter decimal values its really pain to do so (entering value like `0.01` ) I guess we have encountered this before and its tied to the box format (type of data) need to be changed I guess...

Thanks @LiborBatek - was able to reproduce, created an issue on frontend: https://github.com/ynput/ayon-frontend/issues/683

Issue #16, opened 2024-08-01
Render Submission failed in Deadline


moonyuet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When we do farm render, deadline errors out in spite of image files being created in the correct directory.

### Expected Behavior:

Deadline should render the outputs successfully and they would be published afterwards,

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Create Render Instance
2. Publish

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
=======================================================
Error
=======================================================
Error: Renderer returned non-zero error code, 1. Check the log for more information.
   at Deadline.Plugins.PluginWrapper.RenderTasks(Task task, String& outMessage, AbortLevel& abortLevel)

=======================================================
Type
=======================================================
RenderPluginException

=======================================================
Stack Trace
=======================================================
   at Deadline.Plugins.SandboxedPlugin.d(DeadlineMessage bgq, CancellationToken bgr)
   at Deadline.Plugins.SandboxedPlugin.RenderTask(Task task, CancellationToken cancellationToken)
   at Deadline.Slaves.SlaveRenderThread.c(TaskLogWriter ajv, CancellationToken ajw)

=======================================================
Log
=======================================================
2024-08-01 13:16:26:  0: Loading Job's Plugin timeout is Disabled
2024-08-01 13:16:26:  0: SandboxedPlugin: Render Job As User disabled, running as current user 'Kayla'
2024-08-01 13:16:27:  0: Executing plugin command of type 'Initialize Plugin'
2024-08-01 13:16:27:  0: INFO: Executing plugin script 'C:\ProgramData\Thinkbox\Deadline10\workers\DESKTOP-1U8VP0C\plugins\66ab1a1a3d668205295dcca5\Blender.py'
2024-08-01 13:16:28:  0: INFO: Plugin execution sandbox using Python version 3
2024-08-01 13:16:28:  0: INFO: About: Blender Plugin for Deadline
2024-08-01 13:16:28:  0: INFO: The job's environment will be merged with the current environment before rendering
2024-08-01 13:16:28:  0: Done executing plugin command of type 'Initialize Plugin'
2024-08-01 13:16:28:  0: Start Job timeout is disabled.
2024-08-01 13:16:28:  0: Task timeout is disabled.
[...]

PR #17, opened 2024-08-01
Add a check on scene render filepath and repair action in Validate Render Output for Deadline

moonyuet:
## Changelog Description
Resolve https://github.com/ynput/ayon-blender/issues/16

## Additional info
I also refactored the validator a little bit so that it won't always error out.

## Testing notes:

1. Launch Blender
2. Create Render Instance
3. Publish
4. If validate render output for deadline errors out, make sure you perform repair action
5. Publish
6. It should be rendered successfully.

### DIFF
--- a/client/ayon_blender/plugins/publish/validate_deadline_publish.py
+++ b/client/ayon_blender/plugins/publish/validate_deadline_publish.py
@@ -31,6 +31,22 @@ class ValidateDeadlinePublish(
         if not self.is_active(instance.data):
             return
 
+        invalid = self.get_invalid(instance)
+        if invalid:
+            bullet_point_invalid_statement = "\n".join(
+                "- {}".format(err) for err in invalid
+            )
+            report = (
+                "Render Output has invalid values(s).\n\n"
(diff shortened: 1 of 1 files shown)


client/ayon_blender/plugins/publish/validate_deadline_publish.py
```diff
@@ -41,21 +57,30 @@ def process(self, instance):
...
+            )
+            invalid.append(msg)
+        return invalid
 
     @classmethod
     def repair(cls, instance):
         container = instance.data["transientData"]["instance_node"]
         prepare_rendering(container)
         bpy.ops.wm.save_as_mainfile(filepath=bpy.data.filepath)
+        bpy.context.scene.render.filepath = "/tmp\\"
```
BigRoy, line 85: Why do we need the double backslash at the end? That won't work on UNIX devices? This may need to be:
```suggestion
        bpy.context.scene.render.filepath = "/tmp/"
```

LiborBatek (approved):
All working fine and without any issues...been fully tested by rendering on DL and checked integrated data on the Ayon

image

PR #18, opened 2024-08-05
Alembic loaders update the libpath without the removal of the cache modifier

moonyuet:
## Changelog Description
This PR is to make sure the libpath is still updated in the cache modifier without removing the modifier.

## Additional info
We need to discuss if the cache modifiers should be implemented in the load function as there is still issue updating the libpath when the users uses set version to the loaded assets for the first time.

## Testing notes:

1. Launch Blender
2. Load Pointcache/model/camera with alembic loader
3. Set version with scene inventory

### DIFF
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -46,6 +46,8 @@ class CacheModelLoader(plugin.BlenderLoader):
                     if constraint.cache_file.name == prev_filename:
                         constraint.cache_file.name = os.path.basename(libpath)
                     constraint.cache_file.filepath = libpath.as_posix()
+                    constraint.object_path = (
+                        constraint.cache_file.object_paths[0].path)
                     found = True
                     break
             if not found:
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -43,9 +43,9 @@ class CacheModelLoader(plugin.BlenderLoader):
             found = False
             for constraint in obj.constraints:
                 if constraint.type == "TRANSFORM_CACHE":
-                    constraint.cache_file.filepath = libpath.as_posix()
                     if constraint.cache_file.name == prev_filename:
                         constraint.cache_file.name = os.path.basename(libpath)
+                    constraint.cache_file.filepath = libpath.as_posix()
                     found = True
                     break
             if not found:
@@ -55,6 +55,8 @@ class CacheModelLoader(plugin.BlenderLoader):
                 constraint = obj.constraints.new("TRANSFORM_CACHE")
                 bpy.ops.cachefile.open(filepath=libpath.as_posix())
                 constraint.cache_file = bpy.data.cache_files[-1]
+                if constraint.cache_file.name == prev_filename:
+                    constraint.cache_file.name = os.path.basename(libpath)
                 constraint.cache_file.filepath = libpath.as_posix()
                 constraint.cache_file.name = os.path.basename(libpath)
                 constraint.cache_file.scale = 1.0
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -40,28 +40,30 @@ class CacheModelLoader(plugin.BlenderLoader):
         to update the filepath of the alembic.
         """
         for obj in asset_group.children:
-            found = False
             names = [modifier.name for modifier in obj.modifiers
                      if modifier.type == "MESH_SEQUENCE_CACHE"]
+            file_list = [file for file in bpy.data.cache_files
+                         if file.name == prev_filename]
             if names:
                 for name in names:
                     obj.modifiers.remove(obj.modifiers.get(name))
+            if file_list:
+                bpy.data.batch_remove(file_list)
 
-            if not found:
-                # This is to keep compatibility with cameras loaded with
-                # the old loader
-                # Create a new constraint for the cache file
-                modifier = obj.modifiers.new(
-                    name='MeshSequenceCache', type='MESH_SEQUENCE_CACHE')
-                bpy.ops.cachefile.open(filepath=libpath.as_posix())
-                modifier.cache_file = bpy.data.cache_files[-1]
-                modifier.cache_file.name = os.path.basename(libpath.as_posix())
-                modifier.cache_file.filepath = libpath.as_posix()
-                modifier.cache_file.scale = 1.0
-                bpy.context.evaluated_depsgraph_get()
+            # This is to keep compatibility with cameras loaded with
+            # the old loader
+            # Create a new constraint for the cache file
+            modifier = obj.modifiers.new(
+                name='MeshSequenceCache', type='MESH_SEQUENCE_CACHE')
+            bpy.ops.cachefile.open(filepath=libpath.as_posix())
+            modifier.cache_file = bpy.data.cache_files[-1]
+            modifier.cache_file.name = os.path.basename(libpath.as_posix())
+            modifier.cache_file.filepath = libpath.as_posix()
+            modifier.cache_file.scale = 1.0
+            bpy.context.evaluated_depsgraph_get()
 
-                modifier.object_path = (
-                    modifier.cache_file.object_paths[0].path)
+            modifier.object_path = (
+                modifier.cache_file.object_paths[0].path)
 
         return libpath
 
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -40,7 +40,9 @@ class CacheModelLoader(plugin.BlenderLoader):
         to update the filepath of the alembic.
         """
         bpy.ops.cachefile.open(filepath=libpath.as_posix())
-        for i, obj in enumerate(asset_group.children):
+        for obj in asset_group.children:
+            asset_name = obj.name.rsplit(":", 1)[-1]
+            print(asset_name)
             names = [modifier.name for modifier in obj.modifiers
                      if modifier.type == "MESH_SEQUENCE_CACHE"]
             file_list = [file for file in bpy.data.cache_files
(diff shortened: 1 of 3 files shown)


LiborBatek (changes requested):
Im getting errors when trying switch version of my ```Alembic``` product loaded...

here is scrngrab of the console:

image

and how it looks in UI when looking into modifier tab:

image

I did tried 2 different assets and still the issue...

LiborBatek:
The weird thing is that I cant switch even manually (redirecting to different file version)...I will try in different ```blender``` release tho and will inform you.

LiborBatek (commented):
Ok so I have rather used created addon package for blender and also setup a newest ```blender 4.2.0 LTS``` version...now it seems working correctly at least speaking of ```animation``` product type (see vid below)
https://github.com/user-attachments/assets/576a85e5-de2e-4576-b55a-4b38ac7f5e6d

but it seems there is still something fishy speaking of ABC for cameras...it works but not everytime and for every version (need to investigate more what is the pattern for it)

Im also surprised that when assigning some shaders to the alembic cache it looses such edits when switching the filepath to different Alembic file...as I was expecting something else...not sure why

LiborBatek:
Im getting this kind of error in console when loading ```pointcache (ABC)``` not sure if somewhat wrong or so called normal :)
```
Alembic import C:\projects\Ayon_Tests\blender\b3d_shots\sh010\publish\pointcache\pointcacheMain\v014\ATS_sh010_pointcacheMain_v014.abc took 17.98 ms
Alembic import took 18.24 ms
Done
  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\ops.py", line 194, in _process_app_events
    manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'PySide6.QtWidgets.QApplication' object has no attribute 'get_window'
execute_in_main_thread
execute_in_main_thread
```

moonyuet:
> Im getting this kind of error in console when loading `pointcache (ABC)` not sure if somewhat wrong or so called normal :)
> 
> ```
> Alembic import C:\projects\Ayon_Tests\blender\b3d_shots\sh010\publish\pointcache\pointcacheMain\v014\ATS_sh010_pointcacheMain_v014.abc took 17.98 ms
> Alembic import took 18.24 ms
> Done
> Traceback (most recent call last):
>   File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\ops.py", line 194, in _process_app_events
>     manager = GlobalClass.app.get_window("WM_OT_avalon_manager")
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'PySide6.QtWidgets.QApplication' object has no attribute 'get_window'
> execute_in_main_thread
> execute_in_main_thread
> ```
> 
> however I dont get any `Alembic` product into my active blender scene and it basically do nothing (Load Action)...

it's not related to this PR but I would still fix it along with this as it affects the updating asset.

LiborBatek:
Im enclosing whats happening in the ```Blender``` when loading ```pointcache``` animation and trying to switch version via ```Manage``` - the file path got updated in the Alembic cache modifier but the model just dissappears . It also doesnt give any error besides the aformentioned one...


https://github.com/user-attachments/assets/e40b725c-2ee1-4a75-b5d8-4f7d4f0393cb

moonyuet:
> Im enclosing whats happening in the `Blender` when loading `pointcache` animation and trying to switch version via `Manage` - the file path got updated in the Alembic cache modifier but the model just dissappears . It also doesnt give any error besides the aformentioned one...
> 
>  Blender_Set_Version_ABC.mp4

This one is fixed with [f0fa08e](https://github.com/ynput/ayon-blender/pull/18/commits/f0fa08e67ff2401e896aeadbd511c9bf57168b63)

LiborBatek (commented):
Still having the issue after ```Set Version```...my animation pointcache just dissappear due to unresolved file...not sure why.

image

LiborBatek (changes requested):
Ok so I am able to perform ```Manage>Set Version``` but there are several flaws atm...

1) We are altering ```File Path``` for the particular ```Alembic``` product version but not ```Object Path``` which stays the same for all of the alembic cache objects resulting into the same geometry occupying the exact same location (as the data streamed are the same mesh)

Note the scrngrab before setting version to ```v004```
Screenshot 2024-08-20 115525

and after performing the set version action:


Screenshot 2024-08-20 115700

after manually tweaking the ```object path``` suddenly all starts to work nicely:


Screenshot 2024-08-20 115923

2) Another issue been that we are mutliplying the data stream unnecessarilly resulting into multiple data blocks which actually could be just single (as before triggering manage>set version only single data stream exists)

Screenshot 2024-08-20 115800

client/ayon_blender/plugins/load/load_cache.py
```diff
@@ -52,15 +53,15 @@ def _update_transform_cache_path(self, asset_group, libpath, prev_filename):
...
-            modifier.cache_file.name = os.path.basename(libpath.as_posix())
+            cache_file_name = os.path.basename(libpath.as_posix())
+            modifier.cache_file.name = cache_file_name
             modifier.cache_file.filepath = libpath.as_posix()
             modifier.cache_file.scale = 1.0
             bpy.context.evaluated_depsgraph_get()
 
             modifier.object_path = (
-                modifier.cache_file.object_paths[0].path)
+                modifier.cache_file.object_paths[i].path)
```
BigRoy, line 64: Something tells me this update will only work if the amount of objects in the alembic file would not change?

moonyuet: it will still change, the object path means the hierarchy object shapes inside the alembic file.

BigRoy: Sure - but if there are more objects in the alembic than in the asset group currently - it wouldn't have those new objects? Or if the are now less objects in the alembic than in the asset group this would error due to it being out of the length of the object paths array?

moonyuet: > Sure - but if there are more objects in the alembic than in the asset group currently - it wouldn't have those new objects? Or if the are now less objects in the alembic than in the asset group this would error due to it being out of the length of the object paths array?

can you think of better way to deal with it? If we always used the first object path it would end up the scenario like [this](https://github.com/ynput/ayon-blender/pull/18#pullrequestreview-2247620541)

BigRoy: The best way I can think of is:

- Before update build a lookup of current scene objects from their object paths in the alembic.
- Then after update:
    - Update those that existed before and still exist.
    - Delete those scene objects that do not exist in the new state after update.
    - Create the scene objects that do exist after the update but didn't exist before.

This has the caveat that you'd lose any local scene changes on nodes that are removed on updates, etc. but at least your scene contents actually match what the alembic contains?

moonyuet: > The best way I can think of is:
> 
> * Before update build a lookup of current scene objects from their object paths in the alembic.
> * Then after update:
>   
>   * Update those that existed before and still exist.
>   * Delete those scene objects that do not exist in the new state after update.
>   * Create the scene objects that do exist after the update but didn't exist before.
> 
> This has the caveat that you'd lose any local scene changes on nodes that are removed on updates, etc. but at least your scene contents actually match what the alembic contains?

I tested with multiple scenarios when setting version to more or less object paths in the next version than in the previous version. It still reads as expected. But we need to be concerned about how to make it right. It's a bit struggling when trying to implement your approach above, knowing that we need to get the object path from the cache file from the modifier, which is inherited from the object. And comparing it with the object paths from the "removed" modifier (Maybe I did something wrong). It could be some limitation of the mesh sequence modifiers, of which we can't get the object path globally.  The snippet below showed how I got the modifier(but still fails to get the correct object path, unlike the current commit):
```python
    def _update_transform_cache_path(self, asset_group, libpath, prev_filename):
        """search and update path in the transform cache modifier
        If there is no transform cache modifier, it will create one
        to update the filepath of the alembic.
        """
        bpy.ops.cachefile.open(filepath=libpath.as_posix())
        old_object_paths = []
        for i, obj in enumerate(asset_group.children):
            names = [modifier.name for modifier in obj.modifiers
                     if modifier.type == "MESH_SEQUENCE_CACHE"]
            file_list = [file for file in bpy.data.cache_files
                         if file.name.startswith(prev_filename)]
            if names:
                for name in names:
                    modifier = obj.modifiers.get(name)
                    if modifier.cache_file is not None:
                        for obj_path in modifier.cache_file.object_paths:
                            obj_path_name = obj_path.path
                            if obj_path_name in old_object_paths:
                                continue
                            old_object_paths.append(obj_path_name)
                    obj.modifiers.remove(modifier)
            if file_list:
                bpy.data.batch_remove(file_list)

            modifier = obj.modifiers.new(
                name='MeshSequenceCache', type='MESH_SEQUENCE_CACHE')
            modifier.cache_file = bpy.data.cache_files[-1]
            cache_file_name = os.path.basename(libpath.as_posix())
            modifier.cache_file.name = cache_file_name
            modifier.cache_file.filepath = libpath.as_posix()
            modifier.cache_file.scale = 1.0
            bpy.context.evaluated_depsgraph_get()

            modifier.object_path = (
                modifier.cache_file.object_paths[i].path)

        return libpath
```

LiborBatek (changes requested):
Now all seems correct at first glance... there is only one data block for Alembic which is great and supposed to be like that, good.
image

The ```Manage>Set Version``` works well too and user can switch versions without any flaws back and forth. All good too.

Now the issue been that we are having wrong ```object path```s for each part of the Alembic (read each object).

For example below the ```undersuit``` took an object path for ```arms``` instead so it looks like arms and not undersuit figure.


Screenshot 2024-08-21 112613

client/ayon_blender/plugins/load/load_cache.py
BigRoy: Admittedly I know too little of blender to quickly help here API wise. If needed I can put in time and see if I can reproduce and help out code-wise. But it may take me some time and I'm not sure if my time is best spent there now, sorry.

LiborBatek (approved):
All working nicely when ```Set Version``` for ```Alembic``` pointcache!

image

LGTM!

PR #20, opened 2024-08-12
Fix transforms on Empty object children not maintained by ABC cache load.

### LINKED ISSUE #2: Blender: Load pointcache loses transformation on top group/empty

- [X] I have searched the existing issues

### Current Behavior:

This logic [here](https://github.com/BigRoy/ayon-core/blob/8549054098355f20296aca0f994c6a253b820377/client/ayon_core/hosts/blender/plugins/load/load_abc.py#L81-L92) makes it so that the objects are re-parented to an 'asset group' but also make it so that any transformation on the loaded top group is lost.

### Expected Behavior:

Data should not be lost, and preferably any transform or data in the published output is retained on load.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Create scene with cube
2. Parent the cube to an empty
3. Move the empty that has the cube as child.
4. Publish the empty as a `pointcacheMain`
5. Load the pointcache

Now the transformation is lost on load.

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

_No response_

### Additional context:

_No response_

robin-ynput:
## Changelog Description
Resolve #2 

These changes consolidate the way imported hierarchy is re-parented in Blender upon cache load. 
Only top nodes of "Empty" objects without any transforms will be skipped.

Published hierarchy:
image

Loaded hierarchy (note how only Empty.002 is removed cause unecessary):
image

## Additional info
This is my first MR for the project, please do let me know if any troubles.

## Testing notes:

1. Create scene with meshes 
2. Parent some of the meshes to new empty objects
3. Apply random transform on both empty and mesh objects
4. Publish the empty as a pointcacheMain
5. Load the pointcache
6. Ensure all of the transforms are preserved

### DIFF
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -69,24 +69,13 @@ class CacheModelLoader(plugin.BlenderLoader):
                 relative_path=relative
             )
 
-        objects = []
-        uninspected_top_objs = [obj for obj in lib.get_selection() if not obj.parent]
-
-        while uninspected_top_objs:
-            obj = uninspected_top_objs.pop()
-
-            # Ignore EMPTYs without transforms/parent (blank containers)
-            if obj.type == "EMPTY" and obj.matrix_local.is_identity:
-                uninspected_top_objs.extend(obj.children)
-                bpy.data.objects.remove(obj)
-
-            # Reparent top object to asset_group
-            else:
-                obj.parent = asset_group
-                objects.append(obj)
-                objects.extend(obj.children_recursive)
+        objects = lib.get_selection()
 
         for obj in objects:
+            # reparent top object to asset_group
+            if not obj.parent:
+                obj.parent = asset_group
+
             # Unlink the object from all collections
             collections = obj.users_collection
             for collection in collections:
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -69,31 +69,22 @@ class CacheModelLoader(plugin.BlenderLoader):
                 relative_path=relative
             )
 
-        imported = lib.get_selection()
-
-        # Use first EMPTY without parent as container
-        container = next(
-            (obj for obj in imported
-             if obj.type == "EMPTY" and not obj.parent),
-            None
-        )
-
         objects = []
-        if container:
-            nodes = list(container.children)
+        uninspected_top_objs = [obj for obj in lib.get_selection() if not obj.parent]
 
-            for obj in nodes:
+        while uninspected_top_objs:
+            obj = uninspected_top_objs.pop()
+
+            # Ignore EMPTYs without transforms/parent (blank containers)
+            if obj.type == "EMPTY" and obj.matrix_local.is_identity:
+                uninspected_top_objs.extend(obj.children)
+                bpy.data.objects.remove(obj)
+
+            # Reparent top object to asset_group
+            else:
                 obj.parent = asset_group
-
-            bpy.data.objects.remove(container)
-
-            objects.extend(nodes)
-            for obj in nodes:
+                objects.append(obj)
                 objects.extend(obj.children_recursive)
-        else:
-            for obj in imported:
-                obj.parent = asset_group
-            objects = imported
 
         for obj in objects:
             # Unlink the object from all collections


BigRoy:
> Only top nodes of "Empty" objects without any transforms will be skipped.

Any reason we are doing that? Shouldn't we just load whatever data comes in, whatever the Alembic file hold

robin-ynput:
> > Only top nodes of "Empty" objects without any transforms will be skipped.
> 
> Any reason we are doing that? Shouldn't we just load whatever data comes in, whatever the Alembic file hold

Good question, it was there initially I assumed that was by design to avoid complex hierarchy in the outliner. 
@BigRoy not sure who to ask, do you know who could have final call on this ?

client/ayon_blender/plugins/load/load_cache.py
```diff
@@ -69,31 +69,22 @@ def _process(self, libpath, asset_group, group_name):
...
-        )
-
         objects = []
-        if container:
-            nodes = list(container.children)
+        uninspected_top_objs = [obj for obj in lib.get_selection() if not obj.parent]
 
-            for obj in nodes:
-                obj.parent = asset_group
+        while uninspected_top_objs:
```
moonyuet, line 75: Is there any reason why we use while loop instead of for loop, knowing that we can iterate through `uninsepcted_top_objs` ?
We can do something like this:
```python
        uninspected_top_objs = [obj for obj in lib.get_selection() if not obj.parent]

        for obj in uninspected_top_objs:
            # Ignore EMPTYs without transforms/parent (blank containers)
            if obj.type == "EMPTY" and obj.matrix_local.is_identity:
                uninspected_top_objs.extend(obj.children)
                bpy.data.objects.remove(obj)
            # Reparent top object to asset_group
            else:
                obj.parent = asset_group
                objects.append(obj)
                objects.extend(obj.children_recursive)
```

moonyuet:
> > > Only top nodes of "Empty" objects without any transforms will be skipped.
> > 
> > 
> > Any reason we are doing that? Shouldn't we just load whatever data comes in, whatever the Alembic file hold
> 
> Good question, it was there initially I assumed that was by design to avoid complex hierarchy in the outliner. @BigRoy, @moonyuet not sure who to ask, do you know who could have final call on this ?

I dont have really the final call.

Guess what @BigRoy means, we can simply load the alembic data , and add and link to the collection, no need to remove the empty(aka the locator). And it is actually doable by removing some conditions inside the code.

client/ayon_blender/plugins/load/load_cache.py
robin-ynput: You're totally right, I guess I confused myself with the `.pop` thingie and modifying the list while iterating.
Simply re-parenting as it comes makes the code even simpler.

robin-ynput:
> > > > Only top nodes of "Empty" objects without any transforms will be skipped.
> > > 
> > > 
> > > Any reason we are doing that? Shouldn't we just load whatever data comes in, whatever the Alembic file hold
> > 
> > 
> > Good question, it was there initially I assumed that was by design to avoid complex hierarchy in the outliner. @BigRoy, @moonyuet not sure who to ask, do you know who could have final call on this ?
> 
> I dont have really the final call.
> 
> Guess what @BigRoy means, we can simply load the alembic data , and add and link to the collection, no need to remove the empty(aka the locator). And it is actually doable by removing some conditions inside the code.

Perfect then, well I've gone ahead and implemented @BigRoy suggestion (keep as it come), even simpler logic.
Let me know if that works for you guys.

LiborBatek (changes requested):
I have tried to publish some objects and empty with child object and while manually importing as Alembic works (using native b3d File>Import Alembic)

When using AYON > Load it fails and doesnt represent the Alembic correctly...

first scrnshot being correct:
Screenshot 2024-08-14 103034

and the Ayon loaded faulty one...see outliner too


Screenshot 2024-08-14 103104

I still dont get what we are trying to achieve here and I agree with @BigRoy  that we should just take the selection make alembic from it and load it without any "optimalizations"...

Where is the issue in current integration regarding unnecessary empties?

Thx for your effort tho! Cheers!

LiborBatek:
hmm weird that the wrong load happens also when using the latest ```ayon-blender 0.2.2``` addon....

when inspecting the loaded ```pointcache``` that empty acting as a parent for that cube got reset transforms to XYZ > 0 0 0

Anyone any idea whats up??

LiborBatek (approved):

PR #21, opened 2024-08-13
Bugfix/add tmp dir for deadline submission

moonyuet:
## Changelog Description
Add temp directory for deadline submission so that it would not error out during deadline submission.
Continuity of https://github.com/ynput/ayon-blender/pull/17
*Support different system platform

## Additional info
n/a

## Testing notes:

1. Launch Blender
2. Create Render Instance
3. Publish
4. If validate render output for deadline errors out, make sure you perform repair action
5. Publish
6. It should be rendered successfully.

### DIFF
--- a/client/ayon_blender/api/render_lib.py
+++ b/client/ayon_blender/api/render_lib.py
@@ -348,7 +348,7 @@ def prepare_rendering(asset_group):
     # Clear the render filepath, so that the output is handled only by the
     # output node in the compositor.
     bpy.context.scene.render.filepath = (
-        "/tmp\\"if sys.platform == "windows" else "/tmp/"
+        "/tmp\\"if sys.platform == "windows" else ""
     )
     render_settings = {
         "render_folder": render_folder,
--- a/client/ayon_blender/api/render_lib.py
+++ b/client/ayon_blender/api/render_lib.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-
+import sys
 import bpy
 
 from ayon_core.settings import get_project_settings
@@ -347,8 +347,9 @@ def prepare_rendering(asset_group):
 
     # Clear the render filepath, so that the output is handled only by the
     # output node in the compositor.
-    bpy.context.scene.render.filepath = ""
-
+    bpy.context.scene.render.filepath = (
+        "/tmp\\"if sys.platform == "windows" else "/tmp/"
+    )
     render_settings = {
         "render_folder": render_folder,
         "aov_separator": aov_sep,
--- a/client/ayon_blender/plugins/publish/validate_deadline_publish.py
+++ b/client/ayon_blender/plugins/publish/validate_deadline_publish.py
@@ -82,5 +82,4 @@ class ValidateDeadlinePublish(
         container = instance.data["transientData"]["instance_node"]
         prepare_rendering(container)
         bpy.ops.wm.save_as_mainfile(filepath=bpy.data.filepath)
-        bpy.context.scene.render.filepath = "/tmp/"
         cls.log.debug("Reset the render output folder...")


client/ayon_blender/api/render_lib.py
```diff
@@ -347,8 +347,9 @@ def prepare_rendering(asset_group):
 
     # Clear the render filepath, so that the output is handled only by the
     # output node in the compositor.
-    bpy.context.scene.render.filepath = ""
-
+    bpy.context.scene.render.filepath = (
```
kalisp, line 350: This BigRoy question is still valid https://github.com/ynput/ayon-blender/pull/17#discussion_r1699958299

moonyuet: Indeed, the empty path for `bpy.context.scene.render.filepath` is working for some cases. So guess we need to investigate deadline submission issue(e.g. some permission denied eror etc.)
I get inspiration from this website: https://blender.stackexchange.com/questions/119851/render-stops-after-1-frame-render-error-permission-denied-cannot-save-0001

moonyuet:
@BigRoy, we gonna dismiss the use of this output setting as we discussed in the meeting, so are we gonna close this PR or?

BigRoy:
> @BigRoy, we gonna dismiss the use of this output setting as we discussed in the meeting, so are we gonna close this PR or?

I'd say that setting it to `tmp` at least isn't the _correct fix_. We will need to find better ways. Unfortunately it seems that Blender ALWAYS renders to the Output destination set there - and when e.g. `/tmp\` is set it just means that on Windows it renders to `C:/tmp` (not that this is NOT `%TEMP%`). Preferably we're able to completely disable the output and just rely on the compositor outputs instead - but I haven't found a way in Blender to do so.

So, the only remaining alternative is putting the files somewhere that makes sense in production. Like, somehow enforce it to the correct temp dir locally like `%TEMP%` or store them inside the project somewhere as `{work_dir}/renders/tmp` or alike.

LiborBatek (changes requested):
It correctly kicks in the validator when output dir left blank (empty) as seen below:

Screenshot 2024-12-11 111851

but once hitting ```Repair``` it triggers error and do not perform repair action:

```
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.0.11\ayon_core\pipeline\publish\publish_plugins.py", line 350, in process
    plugin.repair(instance)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.6+dev\ayon_blender\plugins\publish\validate_deadline_publish.py", line 83, in repair
    prepare_rendering(container)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.6+dev\ayon_blender\api\render_lib.py", line 350, in prepare_rendering
    bpy.context.scene.render.filepath = temp_render_dir
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: bpy_struct: item.attr = val: RenderSettings.filepath doesn't support None from string types
```

LiborBatek (changes requested):
Ok so now the ```Repair``` action works without any errors! 

There is one caveat with temp folder path which had been set like this:

Screenshot 2024-12-11 134242

```C:\projects\Blender\Shots\sh010\work\lighting\renders\tmp``` which leads to a staging renders been in incorrect place sitting next to that tmp folder as follows:


Screenshot 2024-12-11 134618

I guess its because we miss the last slash in the path so it probably should look more like this instead:

```C:\projects\Blender\Shots\sh010\work\lighting\renders\tmp\``` but thats just me guessing.

Maybe also including host name within the folder path could be preferable so it would end up below ```lighting\renders\blender\tmp\``` again not sure if possible.

**Review by LiborBatek — APPROVED** [review]
Now all behaves correctly!

LGTM

side note: it seems that by filling up the output directory, Blender produces redundant rendered frames in that ```tmp``` folder and those arent cleaned up after ```publish job``` been finnished. Thats something we should think about in future and how to deal with it... as its a kind of blender specific side product outside AYON templating system...

File: client/ayon_blender/plugins/publish/validate_deadline_publish.py [hunk 2]
```diff
@@ -82,5 +82,4 @@ def repair(cls, instance):
         container = instance.data["transientData"]["instance_node"]
         prepare_rendering(container)
         bpy.ops.wm.save_as_mainfile(filepath=bpy.data.filepath)
-        bpy.context.scene.render.filepath = "/tmp/"
-        cls.log.debug("Reset the render output folder...")
+        cls.log.debug("Reset the render output folder...")
```
BigRoy, line 85: This logged message is now in a very odd place.
```suggestion
```

BigRoy:
Note that the changes in https://github.com/ynput/ayon-blender/pull/21/commits/9b0bba3538494fab8291f37f216844ba2ce3cada do not make it cross-compatible on the farm OR in some case even cross-compatible between machines. It is totally valid to override `TMPDIR` environment variable per machine if e.g. it has some storage location you want to set it to instead... which may be a location not available on the other workstations running on the farm.

moonyuet:
> Note that the changes in [9b0bba3](https://github.com/ynput/ayon-blender/commit/9b0bba3538494fab8291f37f216844ba2ce3cada) do not make it cross-compatible on the farm OR in some case even cross-compatible between machines. It is totally valid to override `TMPDIR` environment variable per machine if e.g. it has some storage location you want to set it to instead... which may be a location not available on the other workstations running on the farm.
> 
> Basically this HARDCODES the path on publish to the temp directory location for that particular machine.

Guess it would be much safer for us to revert the change and store the path into the workdir/renders instead? And we can think about how to clean up that tmp directory lasted in workdir later?

LiborBatek (changes requested):
Even when triggering the ```Repair``` action it doesnt fill up the out dir and keep it empty...

when inspecting the console there is some traceback as follows:

```
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.7+dev\ayon_blender\plugins\publish\validate_deadline_publish.py", line 62, in process
    raise PublishValidationError(
ayon_core.pipeline.publish.publish_plugins.PublishValidationError: Render Output has invalid values(s).
```

LiborBatek:
now the ```Repair``` action works and set my out dir to e.g.
```C:/Users/lbate/AppData/Local/Temp/renders/tmp/```

so no errors thrown this time...been tested in blender ```4.3.1``` and ```3.6.5```

My only question been what happens if such folder been set when on the Deadline farm...as probably such directory wont be available?? thats my guessing Im not sure...but good to think about it beforehand

Shouldnt we put something else in the folder path then?? e.g. ```/tmp/``` instead? @moonyuet @BigRoy ?

moonyuet:
> now the `Repair` action works and set my out dir to e.g. `C:/Users/lbate/AppData/Local/Temp/renders/tmp/`
> 
> so no errors thrown this time...been tested in blender `4.3.1` and `3.6.5`
> 
> My only question been what happens if such folder been set when on the Deadline farm...as probably such directory wont be available?? (different machine and users) thats my guessing Im not sure...but good to think about it beforehand
> 
> Shouldnt we put something else in the folder path then?? e.g. `/tmp/` instead? @moonyuet @BigRoy ?

I am thinking if we should create the tmp path inside the work directory instead, this should be avoiding the conflict between the temp and the staging

LiborBatek (approved):
image

LGTM

PR #26, opened 2024-09-04
Fix set frame range and set resolution

BigRoy:
## Changelog Description

Entities were passed but the `attrib` key wasn't accessed. Improved variable names for clarity and fixed the bug.

## Testing notes:

1. Set resolution should work
2. Set frame range should work
3. Both should also be set on new scene (when enabled settings)


moonyuet (approved):
Code looks good for me, also did the simple test in my side with blender. Looks good.

LiborBatek (approved):
Both ```Set Frame Range``` and ```Set Resolution``` works well.
image

Also initial values for those matches when opening the workfile for particular context for a first time!

I guess good to go!

Issue #27, opened 2024-09-05
Allow for including handles in frame range for Blender


jrsndl:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Blender integration now ignores handles and sets the frame range of the work file just to start and end frame.
It would be great to  have  a way to work on handles in a similar fashion to Maya "Include/Exclude handles..." settings.
![image](https://github.com/user-attachments/assets/4d17dd5c-d586-4beb-a4a0-53fba86333c7)

### How would you imagine the implementation of the feature?

Just adding the switch for including the handles. Default to be off, so current behavior is kept.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

BigRoy:
@jrsndl Assigned it to you since I believe you were interested in tackling this yourself.

jrsndl:
> @jrsndl Assigned it to you since I believe you were interested in tackling this yourself.

Sorry, I accidentaly unassigned myself. Can you please assign me again? Thank you

PR #28, opened 2024-09-05
enhancement/AY-6116-blender_allow_handles_frame_range

jrsndl:
## Changelog Description

Allows to configure in settings whether a workfile frame range should include the handles or not for certain tasks. The default is like before where handles are excluded from the timeline.

## Additional info

Should address the issue #27 

Matches a similar feature in `ayon-maya`.

## Testing notes:

1. Upload new addon
2. Tweak the settings for Including handles or not
3. Trigger set frame range


BigRoy:
Resolved the conflicts, pushed some cosmetic fixes and updated the PR description and added reviewers. Happy testing @LiborBatek 🫶

moonyuet (commented):
When I enable the `include_handles_default`, the setting works as expected. But setting up the handle does not work when the task type being enabled,

When the `include_handles_default` settings is disabled, and with the tasktype filtering system set up, all the handles in all task types are not set up by clicking the set frame range.

LiborBatek (approved):
Been tested in both settings ```with handles``` used and when disabled and not used...

image


Works like charm!

BigRoy:
@LiborBatek how about the case that Kayla mentioned:

> When I enable the include_handles_default, the setting works as expected. But setting up the handle does not work when the task type being enabled,
> 
> When the include_handles_default settings is disabled, and with the tasktype filtering system set up, all the handles in all task types are not set up by clicking the set frame range.

It sounds like @moonyuet says the task type filtering does not work as intended, or?

BigRoy:
I had a look at the [Maya implementation here](https://github.com/ynput/ayon-maya/blob/fc52bb1eb3c1d5c5e581f781e76a2aa5871d8634/client/ayon_maya/api/lib.py#L2530) and I think I fixed it with https://github.com/ynput/ayon-blender/pull/28/commits/0d7d2e5190721aa46912e9e1e24fe0864850f0ca @moonyuet could you confirm?

server/settings/include_handles.py
```diff
@@ -0,0 +1,30 @@
+from ayon_server.settings import (
+    BaseSettingsModel,
+    SettingsField,
+    task_types_enum,
+)
+
+
+class IncludeByTaskTypeModel(BaseSettingsModel):
+    task_type: list[str] = SettingsField(
```
iLLiCiTiT, line 9:
```suggestion
    task_types: list[str] = SettingsField(
```

server/settings/include_handles.py
```diff
@@ -0,0 +1,30 @@
...
+        enum_resolver=task_types_enum
+    )
+    include_handles: bool = SettingsField(True, title="Include handles")
+
+
+class IncludeHandlesModel(BaseSettingsModel):
+    include_handles_default: bool = SettingsField(
+        True, title="Include handles by default"
+    )
+    per_task_type: list[IncludeByTaskTypeModel] = SettingsField(
```
iLLiCiTiT, line 21: Couldn't this be profiles?

server/settings/include_handles.py
```diff
@@ -0,0 +1,30 @@
+from ayon_server.settings import (
```
iLLiCiTiT, line 1: Is it so important it deservers separate file? 🙂

server/settings/include_handles.py
```diff
@@ -0,0 +1,30 @@
...
+        default_factory=list,
+        title="Task types",
+        enum_resolver=task_types_enum
+    )
+    include_handles: bool = SettingsField(True, title="Include handles")
+
+
+class IncludeHandlesModel(BaseSettingsModel):
+    include_handles_default: bool = SettingsField(
+        True, title="Include handles by default"
```
iLLiCiTiT, line 19:
```suggestion
        False, title="Include handles by default"
```

server/settings/include_handles.py
BigRoy: Done with https://github.com/ynput/ayon-blender/pull/28/commits/6534f1c5f390359f36ec3ed1bb8e0b0560ce4855

LiborBatek (changes requested):
Yeah  I have made test before the ```Task Profiles``` and it was working fine as Kayla mentioned...

but when ```Profiles``` used it does not take handles into account, need to be addressed

LiborBatek (approved):
I have re-test it again and it works well now!
thanks goes to @iLLiCiTiT pushing (me) to redo it

PR #38, opened 2024-09-17
Add load to compositor loader

### LINKED ISSUE #32: Compositor: Incorporate Loader for Image products

- [X] I have searched the existing issues.

We are missing a loader to load renders/images to the blender compositor.

### How would you imagine the implementation of the feature?

Implement a Load to Compositor loader.

- [X] I have added the relevant labels to the enhancement request.

_No response_

### Additional context:

Requested on [discord](https://discord.com/channels/517362899170230292/1285004215290298388)

Some API references:
- maybe this? https://blender.stackexchange.com/a/175301
- unanswered: https://blender.stackexchange.com/questions/185116/how-can-i-add-an-image-sequence-in-the-compositor-with-python

BigRoy:
## Changelog Description

Add a loader action to load media to the blender compositor in a managed way.

## Additional info

Fixes #32 

image

image

image

## Testing notes:

- [x] Test loading of multiple media types, like sequences, videos, stills, etc.
- [ ] The frame range, colorspace, etc. should be applied as expected from the representation data.
- [x] Updating/removing the loaded containers should work, and do what is expected (e.g. it should not reset unwanted attributes/things in the UI)
- [x] Updating/removing should NOT leave 'redundant' image datablocks in the current files. So it should leave no "Unused data".
- [x] Preferably it doesn't conflict with the "render setup" that also uses compositing nodes?
- [x] Evaluate which product types should be loadable? (Currently it allows `render` only)

1. Load some renders into the compositor

### DIFF
--- a/client/ayon_blender/plugins/load/load_image_compositor.py
+++ b/client/ayon_blender/plugins/load/load_image_compositor.py
@@ -1,0 +1,145 @@
+import os
+from typing import Dict, List, Optional
+
+import bpy
+
+from ayon_core.lib.transcoding import VIDEO_EXTENSIONS
+from ayon_blender.api import plugin, lib
+from ayon_blender.api.pipeline import AVALON_CONTAINER_ID
+
+
+class LoadImageCompositor(plugin.BlenderLoader):
+    """Load media to the compositor."""
+
+    product_types = {"render"}
+    representations = {"*"}
+
+    label = "Load in Compositor"
+    icon = "code-fork"
+    color = "orange"
+
+    def process_asset(
+        self, context: dict, name: str, namespace: Optional[str] = None,
+        options: Optional[Dict] = None
+    ) -> Optional[List]:
+        """
+        Arguments:
+            name: Use pre-defined name
+            namespace: Use pre-defined namespace
+            context: Full parenthood of representation to load
+            options: Additional settings dictionary
+        """
+        path = self.filepath_from_context(context)
+
+        # Enable nodes to ensure they can be loaded
+        if not bpy.context.scene.use_nodes:
+            self.log.info("Enabling 'use nodes' for Compositor")
+            bpy.context.scene.use_nodes = True
+
+        # Load the image in data
+        image = bpy.data.images.load(path, check_existing=True)
+
+        # Get the current scene's compositor node tree
+        node_tree = bpy.context.scene.node_tree
+
+        # Create a new image node
+        img_comp_node = node_tree.nodes.new(type='CompositorNodeImage')
+        img_comp_node.image = image
+        self.set_source_and_colorspace(context, img_comp_node)
+
+        data = {
+            "schema": "openpype:container-2.0",
+            "id": AVALON_CONTAINER_ID,
+            "name": name,
+            "namespace": namespace or '',
+            "loader": str(self.__class__.__name__),
+            "representation": context["representation"]["id"],
+        }
+        lib.imprint(img_comp_node, data)
+
+        return [img_comp_node]
+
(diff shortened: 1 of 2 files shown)


client/ayon_blender/plugins/load/load_image_compositor.py
```diff
@@ -0,0 +1,145 @@
...
+            frame_start_handle = frame_start - handle_start
+            frame_end_handle = frame_end + handle_end
+            duration: int = frame_end_handle - frame_start_handle + 1
+            image_comp_node.frame_duration = duration
+            if source == "SEQUENCE":
+                image_comp_node.frame_start = frame_start_handle
+                image_comp_node.frame_offset = frame_start_handle - 1
+            else:
+                image_comp_node.frame_start = frame_start_handle
+                image_comp_node.frame_offset = 0
```
BigRoy, line 130: Honestly. These were just magical... 

For whatever reason a file sequence with frame start 1001 defaults playback at frame 0, whereas a video at frame start 1001 defaults the playback at frame 1001. Hence, the sequence one requires an offset - and because it does not start at frame 1 but zero it needs to be minus one. 🤷 

It must be that I'm stupid - but if any blender geek can tell me how this _should_ be approach in the compositor to configure a video or sequence to start at the actual frame I want without this that'd be more than welcome.

client/ayon_blender/plugins/load/load_image_compositor.py
```diff
@@ -0,0 +1,150 @@
...
+                image.colorspace_settings.name = colorspace
+
+    def remove_image_if_unused(self, image: bpy.types.Image):
+        if image and not image.users:
+            self.log.debug("Removing unused image: %s", image.name)
+            bpy.data.images.remove(image)
+
+    def switch(self, container, context):
+        # Support switch in scene inventory
+        self.update(container, context)
```
moonyuet, line 150: honestly didn't see the switch function in the blender loaders.  Should we create separate PR to add this back?

LiborBatek (changes requested):
It works nicely for ```render``` products also ```Manage``` tools works well.

image

Been tested with EXRs, PNGs, mp4s all good and also when removed from the workfile no harm caused so far good!

There was one issue tho with ```Color Space``` when using ```PNG``` and ```JPG``` as it was not set to ```sRGB``` but ```ACEScc``` for some reason. However all working fine with ```EXR``` which was correctly set to project color profile ```ACEScg```.

I am also fully in to support also ```Image``` products for loading into Compositor...as its pretty useful (mattepaints from PS etc) - now not supported.

client/ayon_blender/plugins/load/load_image_compositor.py
BigRoy: Moved that to https://github.com/ynput/ayon-blender/pull/43

BigRoy:
> There was one issue tho with Color Space when using PNG and JPG as it was not set to sRGB but ACEScc for some reason. However all working fine with EXR which was correctly set to project color profile ACEScg.

If there is no colorspace data found, I set no colorspace. So this sounds like the default behavior of blender in that case? Not much we can do then i think? Unless we suddenly start to make random assumptions?

> I am also fully in to support also Image products for loading into Compositor...as its pretty useful (mattepaints from PS etc) - now not supported.

Added image and plate. All good now?

LiborBatek:
Thanks, yeah cool you have added those too!

Speaking of color spaces aka file without color space data found...would it make sense to set those non linear ones to sRGB? (PNG, JPG at least) as I think its quite common for those? 

If you think its not good idea, forget it and let it on Blender default behaviour then...

I will make second testing round asap...thx!

PR #42, opened 2024-09-17
Fix optional support for Extract USD Model

BigRoy:
## Changelog Description

Fix optional support for Extract USD Model

image

## Additional info

It was optional by default, and had options in settings but the optional toggle was not shown nor used.

## Testing notes:

1. Optional toggle should show in publisher UI for model product type.
2. The inactive (disabled in publisher UI) state should be adhered to and do no export when toggled off.

### DIFF
--- a/client/ayon_blender/plugins/publish/extract_usd.py
+++ b/client/ayon_blender/plugins/publish/extract_usd.py
@@ -2,11 +2,12 @@ import os
 
 import bpy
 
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
It does work as it should been tested in multiple scenarios (optional, non optional, enabled vs disabled) also checked the published files if the publish produced just proper repre types
...
image

LGTM

PR #44, opened 2024-09-17
Fix frame range attributes for instances + validations

BigRoy:
## Changelog Description

- Discard 'step' for pointcache/animation since it did not influence the output alembic at all.
- Default the frame range to current context task entity instead of current scene
- Add handle start and handle end attributes to be explicit about that, defaulting to current task attribs.
- Actually fix passing along the start and end frame to the Alembic export from creator to instance data.
- Add support for validating frame range against the instance's task/folder entity for review, pointcache, animation, etc. + support repair method for those.
- Remove redundant 'scene overrides' that tried to use that to affect what the alembic export did - but it didn't do much, and the export command has dedicated arguments for that.

## Additional info

I have no idea whether the `step` attribute does work for the other export methods, e.g. FBX or alike but we may need to double check that to see whether the `step` attribute is relevant anywhere.

Sorry for the big changes, but this really was way easier to fix when combining them.

## Testing notes:

The following product types should adhere to the frame ranges specified on the instances, be validated against the frame range (optionally) and when invalidated should allow repair to work.
1. [x] animation
2. [x] camera
3. [x] pointcache
4. [ ] render
5. [x] review

### DIFF
--- a/client/ayon_blender/api/colorspace.py
+++ b/client/ayon_blender/api/colorspace.py
@@ -22,18 +22,20 @@ class RenderProduct(object):
 
 
 class ARenderProduct(object):
-    def __init__(self):
+    def __init__(self, frame_start, frame_end):
         """Constructor."""
         # Initialize
-        self.layer_data = self._get_layer_data()
+        self.layer_data = self._get_layer_data(frame_start, frame_end)
         self.layer_data.products = self.get_render_products()
 
-    def _get_layer_data(self):
-        scene = bpy.context.scene
-
+    def _get_layer_data(
+        self,
+        frame_start: int,
+        frame_end: int
+    ) -> LayerMetadata:
         return LayerMetadata(
-            frameStart=int(scene.frame_start),
-            frameEnd=int(scene.frame_end),
+            frameStart=int(frame_start),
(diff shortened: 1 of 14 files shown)


moonyuet (approved):
Did some simple test. Publish successfully with validate frame range.
image

Tested with incorrect frame range and repair action also works for fixing the frame range.
image

BigRoy:
@LiborBatek any chance you can test the **render**?

LiborBatek (changes requested):
First thanks for such a massive improvement / refactoring!!! Much appreciated...

It works well for most publish instance types!

The only malfunctioning is the ```Render``` as when validating the scene it doesnt take the values on the publish instance but global frame range (timeline at the bottom) this being also true when using ```Repair``` action as it tweaks the global timeline to correct values but not update the values on publish ```Render``` instance...

Seems like it somehow bypassing those instance frame range values atm.
Screenshot 2024-09-19 112647

BigRoy:
Thanks @LiborBatek - should be fixed now. 

There was definitely a lot broken since the merge of #1

LiborBatek (approved):
Now it also works for ```Render``` publish instance...

LGTM!

PR #52, opened 2024-09-30
Make sure there is object path filled into the cache modifier

moonyuet:
## Changelog Description
This PR is the quick fix on the object path not being filled into the mesh sequence modifier.

## Additional info
n/a

## Testing notes:

1. Launch Blender
2. Load PointCache, Model and Animation
3. Update/Setting Version to the geometry
4. Repeat 3
5. Object path of the modifier should be filled.

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -566,28 +566,14 @@ def collect_animation_defs(create_context, step=True, fps=False):
 
 def get_cache_modifiers(obj, modifier_type="MESH_SEQUENCE_CACHE"):
     modifiers_dict = {}
-    modifiers = []
-    modifier_names = [modifier.name for modifier in obj.modifiers
-                      if modifier.type == modifier_type]
-    if modifier_names:
-        modifiers = [modifier for modifier in obj.modifiers
-                     if modifier.type == modifier_type]
-        modifiers_dict = {
-            obj.name: (
-                modifier for modifier in obj.modifiers
-                if modifier.type == modifier_type
-            )
-        }
+    modifiers = [modifier for modifier in obj.modifiers
+                 if modifier.type == modifier_type]
+    if modifiers:
+        modifiers_dict[obj.name] = modifiers
     else:
         for sub_obj in obj.children:
             for ob in sub_obj.children:
                 cache_modifiers = [modifier for modifier in ob.modifiers
                                    if modifier.type == modifier_type]
-                modifiers.extend(cache_modifiers)
-                modifiers_dict.update({
-                    ob.name: (
-                        modifier for modifier in ob.modifiers
-                        if modifier.type == modifier_type
-                    )
-                })
+                modifiers_dict[ob.name] = cache_modifiers
     return modifiers_dict
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -524,3 +524,18 @@ def collect_animation_defs(fps=False):
         defs.append(fps_def)
 
     return defs
+
+
+def get_cache_modifiers(obj, modifier_type="MESH_SEQUENCE_CACHE"):
+    modifiers = []
+    modifier_names = [modifier.name for modifier in obj.modifiers
+                      if modifier.type == modifier_type]
+    if modifier_names:
+        modifiers = [modifier for modifier in obj.modifiers
+                     if modifier.type == modifier_type]
+    else:
+        for sub_obj in obj.children:
+            for ob in sub_obj.children:
+                modifiers = [modifier.name for modifier in ob.modifiers
+                            if modifier.type == modifier_type]
+    return modifiers
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -565,15 +565,29 @@ def collect_animation_defs(create_context, step=True, fps=False):
 
 
 def get_cache_modifiers(obj, modifier_type="MESH_SEQUENCE_CACHE"):
+    modifiers_dict = {}
     modifiers = []
     modifier_names = [modifier.name for modifier in obj.modifiers
                       if modifier.type == modifier_type]
     if modifier_names:
         modifiers = [modifier for modifier in obj.modifiers
                      if modifier.type == modifier_type]
+        modifiers_dict = {
+            obj.name: (
+                modifier for modifier in obj.modifiers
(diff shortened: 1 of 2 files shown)


LiborBatek (changes requested):
Hmm Im getting errors when trying to perform ```Manage > Set Version```
Screenshot 2024-09-30 174308



```

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\ops.py", line 128, in execute
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_cache.py", line 277, in exec_update
    libpath = self._update_transform_cache_path(asset_group, libpath, prev_filename)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_cache.py", line 58, in _update_transform_cache_path
    modifier.cache_file = bpy.data.cache_files[-1]
    ^^^^^^^^^^^^^^^^^^^

AttributeError: 'NoneType' object has no attribute 'cache_file'
```

LiborBatek (changes requested):
Even tho it got updated to latest version it doesnt fill in the updated file path to the actual ABC pointcache

image

No errors thrown.

LiborBatek (changes requested):
Still behaves the same...again the Scene Inventory got new version (once triggering the Update action) but the pointcache modifier lacks file path and been empty.

LiborBatek (changes requested):
Gives me an error when trying to update the version (I have also produced fresh new pointcache products to be sure of compatibility)

```

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\ops.py", line 128, in execute
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_cache.py", line 280, in exec_update
    libpath = self._update_transform_cache_path(asset_group, libpath, prev_filename)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_cache.py", line 59, in _update_transform_cache_path
    if modifier.type == "MESH_SEQUENCE_CACHE":
       ^^^^^^^^^^^^^

AttributeError: 'str' object has no attribute 'type'

```

Screenshot 2024-10-07 143650

LiborBatek:
Now it doesnt show any error but obviously does not finnish the ```Update to latest``` as it fails on first geo as seen on the modifier and object type mismatch info

image

Once I manually point the correct geo it starts to work and uses that pointcache....
image

image

client/ayon_blender/api/lib.py
```diff
@@ -562,3 +562,32 @@ def collect_animation_defs(create_context, step=True, fps=False):
...
+                cache_modifiers = [modifier for modifier in ob.modifiers
+                                   if modifier.type == modifier_type]
+                modifiers.extend(cache_modifiers)
+                modifiers_dict.update({
+                    ob.name: (
+                        modifier for modifier in ob.modifiers
+                        if modifier.type == modifier_type
+                    )
+                })
+    return modifiers_dict
```
BigRoy, line 593: What's going on here? Why doing so many type checks?
And why do we care about specifically the 'children's children' in the else case?

```suggestion
    modifiers_dict = {}
    modifiers = [modifier for modifier in obj.modifiers
                 if modifier.type == modifier_type]
    if modifiers:
        modifiers_dict[obj.name] = modifiers
    else:
        # QUESTION: Why do we care about the obj's children's children?
        for sub_obj in obj.children:
            for ob in sub_obj.children:
                cache_modifiers = [modifier for modifier in ob.modifiers
                                   if modifier.type == modifier_type]
                modifiers_dict[ob.name] = cache_modifiers
    return modifiers_dict
```

Isn't this the same? but again- why the 'children' iterations? Seems off.

moonyuet: I know it doesn't make sense but when I tested with Libor's assets, the modifiers creation lasted inside the geo. The screenshot may explain you more.
image

BigRoy: That's correct behavior - but we should just iterate **all** objects, not a specific objects childrens' children?
Also, there's two types of caches, `MeshSequenceCache` for vertex data and there's a constraint or whatever it was I believe that's present on alembic transformations (that are not vertex changes but animated transforms).

moonyuet: > That's correct behavior - but we should just iterate **all** objects, not a specific objects childrens' children? Also, there's two types of caches, `MeshSequenceCache` for vertex data and there's a constraint or whatever it was I believe that's present on alembic transformations (that are not vertex changes but animated transforms).

you mean the transform constraint? We used it for the alembic camera loader. It used to be there in the cache loader, but it somehow breaks MeshSequenceCache during implementation of both modifiers, so I removed it.

BigRoy: An object _may_ have both, have one of them or none of them. It depends on the animation in the alembic file.

- If the object has no vertex animation (e.g. no skinned deformations) then it will get no alembic mesh modifier
- If the object has no transform animation (e.g. no actual animated transforms) then it will get no constraint modifier.

Let me know if you need me to investigate further

moonyuet: > An object _may_ have both, have one of them or none of them. It depends on the animation in the alembic file.
> 
> * If the object has no vertex animation (e.g. no skinned deformations) then it will get no alembic mesh modifier
> * If the object has no transform animation (e.g. no actual animated transforms) then it will get no constraint modifier.
> 
> Let me know if you need me to investigate further

As mentioned above, we can re-implement and we need to set up the condition where we check on the transform and geometry of the asset_group children, and set up the modifier.

LiborBatek (approved):
Now it works great and I was able to go up version (latest) and also back to older versions...

So both ```Update to latest``` and ```Set Version``` actions produces correct outcome and works nicely!

Screenshot 2024-10-08 152831


Screenshot 2024-10-08 152858

moonyuet:
I have created the github issue for this so that we can track on this later

PR #66, opened 2024-10-31
Add Load image to Shader Editor for Select mesh

BigRoy:
## Changelog Description

Add Load image to Shader Editor for Select mesh

Based on community contribution [here](https://discord.com/channels/517362899170230292/1248298172816822424/1301494113790853212) by Alain Xerri (Atomic). Thanks!

## Testing notes:

1. Select a mesh, use the `Load in Shader Editor to selected mesh` loader action
2. Image should be loaded into its material node tree
3. The image should be managed (can be updated/removed from scene inventory)

### DIFF
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -638,6 +638,21 @@ def ls() -> Iterator:
 
             yield parse_container(node)
 
+    # Shader nodes are not available in a way that `lib.lsattr` can find.
+    for material in bpy.data.materials:
+        material_node_tree = material.node_tree
+        if not material_node_tree:
+            continue
+
+        for shader_node in material_node_tree.nodes:
+            if not shader_node.get(AVALON_PROPERTY):
+                continue
+
+            if shader_node.get(AVALON_PROPERTY).get("id") not in container_ids:
+                continue
+
+            yield parse_container(shader_node)
+
 
 def publish():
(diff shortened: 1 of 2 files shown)


LiborBatek (changes requested):
In general, it works well and without any errors...

Also tested with ```AYON > Manage``` and also works fine when setting versions and removing the product from the workfile...

There is only minor thing and thats  its quite loooong in the actions, we should stick to something short and clear really.

image

Any proposals to make it shorter and more clear?

LiborBatek:
after your commit it looks perfect as the action name is short and clear

image

BigRoy:
@LiborBatek should this be approved - or was there still something todo? Your review is still on request changes.

Only thing I can think of maybe is UDIM / image sequence support? But that _may_ already work? 🗡️

LiborBatek:
@BigRoy sry, I missed that one even tho I have tested it with UDIMs and its behaving a bit weird atm...as they act as separate files so not fully working (even tho its not failing aka erroring out hard to call it functional atm)


Screenshot 2024-10-31 154626

not sure tho how easy is to fix it speaking of blender and UDIMs as I remember it was not very straightforward to se file node to UDIM mode...need to re-check how things are done in the latest b3d release

alainxi:
If I may :  it does not need to be perfect on day one.   UDIM could be added later.

LiborBatek (approved):

PR #67, opened 2024-11-05
Refactor render workflow to manage render compositor network outputs instead of forced render structure

BigRoy:
## Changelog Description

Allow the Compositor's Node Output File to be the ground truth about the "output" of your render.

It should not matter whether it's the direct 3D render or whether you are doing slap-comp compisiting in-between. The idea here is that there just happens to be a node, with some input layers defined, which then end up defining the actual render products you want to publish.

Currently the so called "aov" suffix to the product is defined by the filename of the layer in the compositor node output file - but we could look into other options, like using custom attributes on the node or another explicit naming option for those. The logic is currently in the `get_aov_identifier` function.

- [ ] TODO: Currently the creator 'auto-collects' from the compositor node tree - which is very different of other behavior on publish instances. We should work towards fixing that. :)

## Testing notes:

1. Build up a nice render network via the compositor node tree using the Compositor Node Output File and connecting your layers to it.
2. Go and publish.

### DIFF
--- a/client/ayon_blender/plugins/create/create_render.py
+++ b/client/ayon_blender/plugins/create/create_render.py
@@ -1,15 +1,15 @@
 """Create render."""
 import bpy
 
-from ayon_core.lib import version_up
-from ayon_core.pipeline.context_tools import version_up_current_workfile
+from ayon_core.pipeline import AYON_CONTAINER_ID
+from ayon_core.pipeline.create import CreatedInstance
+
 from ayon_blender.api import plugin, lib
 from ayon_blender.api.render_lib import prepare_rendering
-from ayon_blender.api.workio import save_file
 
 
 class CreateRenderlayer(plugin.BlenderCreator):
-    """Single baked camera."""
+    """Create render instance."""
 
     identifier = "io.openpype.creators.blender.render"
     label = "Render"
@@ -19,6 +19,13 @@ class CreateRenderlayer(plugin.BlenderCreator):
     def create(
         self, product_name: str, instance_data: dict, pre_create_data: dict
     ):
+        # TODO: Create compositor node and imprint there.
+        # Create a compositor `CompositorNodeOutputFile` node as instance node
+        # And pre-connect any existing passes according to settings.
+        # It's important this does not mess up other existing networks in the
+        # compositing node tree, so that it remains non-destructive.
+        # TODO: Do not override scene render engine (unless explicitly enabled)
+
         try:
             # Run parent create method
             collection = super().create(
@@ -31,24 +38,60 @@ class CreateRenderlayer(plugin.BlenderCreator):
             bpy.data.collections.remove(collection)
             raise
 
-        # TODO: this is undesiderable, but it's the only way to be sure that
-        # the file is saved before the render starts.
-        # Blender, by design, doesn't set the file as dirty if modifications
-        # happen by script. So, when creating the instance and setting the
-        # render settings, the file is not marked as dirty. This means that
-        # there is the risk of sending to deadline a file without the right
-        # settings. Even the validator to check that the file is saved will
-        # detect the file as saved, even if it isn't. The only solution for
-        # now it is to force the file to be saved.
-        if not bpy.data.filepath:
-            version_up_current_workfile()
-        else:
-            filepath = version_up(bpy.data.filepath)
-            save_file(filepath, copy=False)
-
         return collection
+
+    def collect_instances(self):
+
+        # Collect regularly
+        super().collect_instances()
+
+        # TODO: Convert legacy instances so that they are imprinted on the
+        #  output node instead.
+        # TODO: Maybe do not auto-collect any compositor node output file
+
+        # Also collect any render output file nodes that are not 'registered'
+        # yet
+        if not bpy.context.scene.use_nodes:
+            return
+
+        tree = bpy.context.scene.node_tree
+        for node in tree.nodes:
+            if node.bl_idname != "CompositorNodeOutputFile":
+                continue
+
+            # If not is already imprinted and hence 'registered' we skip
+            # it to avoid registering twice.
+            project_name = self.create_context.get_current_project_name()
+            folder_entity = self.create_context.get_current_folder_entity()
+            task_entity = self.create_context.get_current_task_entity()
+            task_name = self.create_context.get_current_task_name()
+            host_name = "blender"
+            variant = node.name
+            product_name = self.get_product_name(
+                project_name,
+                folder_entity,
+                task_entity,
+                variant,
+                host_name,
+                project_entity=self.create_context.get_current_project_entity()
+            )
+            data = {
+                "folderPath": folder_entity["path"],
+                "task": task_name,
+                "variant": variant,
+                "productName": product_name,
+                "productType": self.product_type,
+                "creator_identifier": self.identifier,
+                "id": AYON_CONTAINER_ID
+            }
+            instance = CreatedInstance(
+                self.product_type, product_name, data, self
+            )
+            instance.transient_data["instance_node"] = node
+
+            # Add instance to create context
+            self._add_instance_to_context(instance)
 
     def get_instance_attr_defs(self):
         defs = lib.collect_animation_defs(self.create_context)
-
         return defs
--- a/client/ayon_blender/plugins/publish/collect_render.py
+++ b/client/ayon_blender/plugins/publish/collect_render.py
@@ -1,6 +1,3 @@
-# -*- coding: utf-8 -*-
-"""Collect render data."""
-
 import os
 import re
 
@@ -11,7 +8,25 @@ from ayon_blender.api import colorspace, plugin
 
 
 class CollectBlenderRender(plugin.BlenderInstancePlugin):
-    """Gather all publishable render instances."""
+    """Gather all publishable render instances.
+
+    For the instance node (bpy.types.CompositorNodeOutputFile) we collect the
+    configured output paths (FileSlots or LayerSlots) and their colorspaces.
+
+    ### AOV identifiers
+
+    When multiple outputs are present (only the case when not rendering to
+    multilayer EXR) then we assign each output an 'aov identifier' that will
+    be added to the product name. So that product: `renderLightingMain` becomes
+    for example `renderLightingMain.beauty` and `renderLightingMain.diffuse`.
+
+    ### Requires enabled compositing node tree
+
+    The render workflow requires Blender to be configured to use the
+    Compositor Node Tree, because it relies on `CompositorNodeOutputFile` to
+    define the output files for rendering.
+
+    """
 
     order = pyblish.api.CollectorOrder + 0.01
     hosts = ["blender"]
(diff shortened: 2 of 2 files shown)


LiborBatek:
I will give it a shot and see how it performs....

But just by briefly seeing this PR I have to note that the main labour which need to be taken care of is the creation of that ```file output``` compositing node with corresponding node tree connections automatically originating from users selection of different AOVs within blender UI + viewLayer(s) present within blender scene as those are the most painful tasks to tackle manually (something blender lacks heavily compared to maya)  user basically having just 2 options by default in blender

1) to go with multichannel EXR
2) manually hooking all nodes within compositor and using ```File output``` node for producing separate passes (same with render layers - view layers in blender terms)

Our legacy blender integration did that automatically for the user even tho been sort of rigid and blender-addon config driven (which is not desirable now and reasoning behind refactoring)

LiborBatek:
So here are my findings so far:

1) It automatically creates ```file output``` node and connects ```beauty``` and ```composite``` outputs which is nice
2) If any AOVs enabled in ```View Layer Properties``` tab for output - these are not taken into account and got swiped off once ```Render``` instance got created
3) Some weird "Render" instance pops up while working with the blender workfile and re-opening the publisher (see img below) - not sure what that really means
4) Once I try to publish the render it fails with following error:

```
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.4+dev\ayon_blender\plugins\publish\collect_render.py", line 46, in process
    output_paths = self.get_expected_outputs(comp_output_node)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.4+dev\ayon_blender\plugins\publish\collect_render.py", line 136, in get_expected_outputs
    base_path: str = node.base_path
                     ^^^^^^^^^^^^^^
AttributeError: 'Collection' object has no attribute 'base_path'
```

Here is the publisher and kind of redundant render instance?!

 
Screenshot 2024-11-06 113641

BigRoy:
> But just by briefly seeing this PR I have to note that the main labour which need to be taken care of is the creation of that `file output` compositing node with corresponding node tree connections automatically originating from users selection of different AOVs within blender UI + viewLayer(s) present within blender scene as those are the most painful tasks to tackle manually (something blender lacks heavily compared to maya) user basically having just 2 options by default in blender
> 
>     1. to go with multichannel EXR
> 
>     2. manually hooking all nodes within compositor and using `File output` node for producing separate passes (same with render layers - "view layer" in blender terms)
> 
> 
> Our legacy blender integration did that automatically for the user even tho been sort of rigid and blender-addon config driven (which is not desirable now and reasoning behind refactoring)

I want to remark that we should really try and get as much of that out of the AYON settings - and instead allow someone to build that almost like a template inside Blender or alike and then e.g. allow using that as an actual loadable product or as workfile templates, or something along those lines.

We should really allow anyone to create any custom node graph - no matter how complex - without us enforcing some structure from our AYON settings. Of course a simple "creation system" like we have we could preserve, but I think the actual usefulness is so much less than just using _actual_ Blender node content somehow the user can actually author in a blender file itself.

The admin of settings may also not be the artist who knows blender - as such, configuring this "template" in settings is too far off from the artist who just knows blender.

Anyway, thanks for the quick test run on the draft PR. Some of the issues you mentioned I know why they happen and what would be the next step - but it's perfect having them clearly laid out and mentioned here, so thanks for that! ❤️

LiborBatek:
Here is a small breakdown what we need to achieve... when creating ```Render``` publish instance(s)

1) Output Renderlayer ```Beauty``` and any present ```AOV```
2) Output ```Composite``` (optional - so user could disable it if not needed)
3)  Output ```Custom File Output``` nodes  user adds before triggering creation of the ```Render``` publish instance

Screenshot 2024-12-04 095159_b

B3d_Re(https://github.com/user-attachments/assets/7edf07e5-8fd5-4fec-ad4a-7e8479330ccb)



1) and 2) been obvious but speaking of 3) we should look for any already present ```File Output``` nodes with any connections and include those for ```Render``` too - maybe via some custom tags/ naming so they could be taken into account for publishing too??

Also worth mentioning that any user output can originate from any node graph location/AOV so it makes sense to just detect the file output node

![Screenshot 2024-12-04 100444

LiborBatek:
@BigRoy @moonyuet  I still think there is quite a value in current implementation and maybe just briefly checking whats the issue with the additional render layer(s) (which are broken atm) but the rest is functional coould be good idea too....

as if just minor issue tweaking might lead to fully functional rendering from blender before we introduce this refactoring above.

moonyuet:
> * Create a render instances when there is already a render instance/comp file output node - that's where I'm still thinking about how that would work. As far as I understood it doesn't actually make sense in Blender to have multiple File Output Nodes? We can support it and work towards it - I just know too little of Blender to say something worthwhile there.

Blender is allowed to have multiple File Output Nodes for render output. 
Do you think that it's good idea to allow users to create the render instance on new or existing file output nodes but also supports to collect data from certain file output nodes through the setting in Publisher UI? 

Moreover, we can make it more artist-friendly, like creating some templates settings which can be driven in ayon settings,  so that artists can just choose whether they want to do multipart or not etc.

LiborBatek:
> @BigRoy @moonyuet I still think there is quite a value in current implementation and maybe just briefly checking whats the issue with the additional render layer(s) (which are broken atm) but the rest is functional could be good idea too....
> 
> as if just minor issue tweaking might lead to fully functional rendering from blender before we introduce this refactoring above. I can imagine that been caused just by some API change within new Blender release (as the rendering features were introduced when blender 3.6 been released and got broken with blender 4.x)

I have made some tests and it seems that it doesnt fail on multiple ```ViewLayer```s but its not supported atm...as it gives no errors once triggering creation ```Render``` publish instance and any additional ```ViewLayer```s been simply ignored as seen below (2 layers present and just single been taken into account for outputs by AYON)

Screenshot 2024-12-04 114814

LiborBatek:
@BigRoy thx for your breakdown! its all sounds thrilling and awesome! love it already :)
speaking of that "tricky part" I guess wen do need to alter the file output names a lot as users could have some weird naming and not using any versioning so we might need to forcing some proper ones ( e.g. variant, viewlayer, version etc.) and also alter the file path accordingly as those arent dynamically created within blender at all...

does that sound still easy? if yes then perfect!!

LiborBatek:
So far seems working fine (meanwhile before my images got rendered) I have found that the file names do lack ```product type``` and having just ```variant``` and ```ViewLayer``` in the file names... not sure but def differs from e.g. ```maya``` host but not neccessarily an issue I guess

image

LiborBatek:
I think this also interesting (I guess) correct me @BigRoy if irrelevant tho

examining the render Job Submission Params in DL itself: the output and the naming with which seems incorrect and also not sure about that suffix ```beauty``` in there...
Screenshot 2025-06-18 124110

LiborBatek:
@BigRoy I hope its ok to test in official latest blender 4.4.3 not just in that beta 4.5 or??

BigRoy:
> @BigRoy I hope its ok to test in official latest blender 4.4.3 not just in that beta 4.5 or??

Should be - yes. Let me just go through feedback and see what I can do soon to get this off the ground.

LiborBatek:
it seems that issue above was caused by that present file output node...once removed and tried to recreate the ```Render``` instance suddenly all worked ok

image

LiborBatek:
I was also able to publish 2 ViewLayers (layer_FR, layer_BG) to DL but briefly checking metadata.json and rendered images they have a wrong naming:

My compositor setup before publish:
Screenshot 2025-06-19 165335

file output node details:
Screenshot 2025-06-19 165656

and metadata JSON:
Screenshot 2025-06-19 165807

Otherwise the batch finnished and saved to correct location ```.../layer_FR``` and ```.../layer_BG```

I have just tested single batch for rendering not full image sequence

And as been said already it worked only if I hadnt any ```File Output Node``` present before triggering ```Render``` instance creation...

LiborBatek:
I have torture test a bit and maybe we should also take care when for some reason the file output node doesnt have any/valid output filepath e.g. if empty?

Im getting failure when trying to publish submit to DL

Screenshot 2025-06-24 141208




```
DEBUG: Collected output path for AOV <beauty>: .exr####
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2502101448_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.11+dev\ayon_blender\plugins\publish\collect_render.py", line 85, in process
    expected_files[aov_identifier] = self.generate_expected_frames(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.11+dev\ayon_blender\plugins\publish\collect_render.py", line 267, in generate_expected_frames
    raise ValueError(
ValueError: Path '.exr####' does not contain a frame token '#'.
```

Would be nice to have some human readible msg and Repair/Select Invalid action? A bit of edge case but could happen so why not to prepare for such scenario, right?


Other than that....all seems working nicely!

@BigRoy not sure about the logic when creating ```render``` publish instance...does it follow ```ayon-blender``` addons settings below or uses what user has set in advance in his workfile in render globals?? If using from the addons setting preference maybe we should introduce some Enable/Disable knob to not do so....so it would respect the workfile preference instead...does this make sense or is completely off?

  
Screenshot 2025-06-24 141805

moonyuet:
> and one last thing...we should probably change the addons defaults as when 'AOVs to create' left empty, it still does hook up `RGBA` default pass no matter what set in the addons settings... so keep it empty as defaults make sense too...
> 
> didnt spot any difference when `Combined` present or not. Screenshot 2025-06-24 144024
> 
> new defaults?
> 
> Screenshot 2025-06-24 144007
> if used it still works for RGBA:
> 
> Screenshot 2025-06-24 143959
> 
> Thoughts?
We can just leave it as empty(without combined options) as we already have multilayer(exr) options. From the user perspective, combined could mean the combination of all renderpasses or multilayers. So the setting might add more confusions.

client/ayon_blender/plugins/create/create_render.py
```diff
@@ -1,53 +1,107 @@
...
     icon = "eye"
 
+    # TODO: Convert legacy instances to new style instances by finding the
+    #  relevant file output node and moving the imprinted data there.
+
+    def _find_existing_compositor_output_node(self) -> Optional["bpy.types.CompositorNodeOutputFile"]:
+        if not bpy.context.scene.use_nodes:
+            return None
+
+        # TODO: If user has a selected compositor node, prefer that one
```
moonyuet, line 24:
```suggestion
        # TODO: If user has a selected compositor node, prefer that one
```
Should we do it now instead of just forcing one particular node?

BigRoy:
@LiborBatek this may need another testing round when you have time. It should now always detect ANY CompositorNodeOutputFile node (even when not created via the Publisher UI) + creation of render setup is now optional on create render (+ the create render setup is added as option in the AYON menu)...

BigRoy:
Additional requests:

> - [ ] It would be awesome if the Creator: Render - Create Render Setup can have an option to only link to selected Scene Render Layer node(s). That way the great linking feature would still be working, and artist would have a way to control which render layers to link. Now it always links every layer, and cleaning that might be more work that to actually link by hand...
> - [ ] when Creator: Render - Create Render Setup option is off, would it be possible to make the render Path the same as with this option to be on? The default path is now to render/tmp, and has no hashes. Just QoL feature, but might help artists to start smoothly.

client/ayon_blender/plugins/publish/validate_deadline_publish.py
```diff
@@ -66,73 +143,59 @@ def process(self, instance):
...
+            cls.log.warning("No workfile scene filepath set. "
+                            "Please save the workfile.")
+            return invalid
+
+        workfile_filename = os.path.basename(workfile_filepath)
+        workfile_filename_no_ext, _ext = os.path.splitext(workfile_filename)
+        cls.log.debug(
+            f"Found compositor output node '{output_node.name}' "
+            f"with base path: {output_node.base_path}")
+        if workfile_filename_no_ext not in output_node.base_path:
```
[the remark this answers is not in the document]
BigRoy: This is intentional, the workfile scene name may be anywhere in the path if using a custom filepath.

BigRoy:
The listed changes have been implemented, see bottom checklist [here](https://github.com/ynput/ayon-blender/pull/67#issuecomment-3057626244). 

Would be great to test again. Especially test the "multiple render instances should generate one render job in deadline" because I haven't done much testing there. (Note that _that_ requires the `ayon-deadline` PR here: https://github.com/ynput/ayon-deadline/pull/187

BigRoy:
> I have tried with this scenario and guess maybe I did something wrong. I created instance without renderSetup. image And then I hit error for submit publish job in deadline. image
> 
> The attachment is the publish report [publish-report-250714-16-04.json](https://github.com/user-attachments/files/21211582/publish-report-250714-16-04.json)

Thanks @moonyuet - can you try again with latest commit?

BigRoy:
> > Thanks @moonyuet - can you try again with latest commit?
> 
> It published and rendered successfully with blender but fails in submitting publish job when I tried to use existing rendersetups. image
> 
> Enclosed with the report from the deadline [Job_2025-07-14_18-29-52_6874dc20583fd45b9acf4a6e.txt](https://github.com/user-attachments/files/21214101/Job_2025-07-14_18-29-52_6874dc20583fd45b9acf4a6e.txt)
> 
> image

My gut feeling here is that you have multiple instances writing into the same **folder** using the same filenames perhaps?
Anyway, can you pass the metadata json files please?
These ones:
```
['renderCompositingFileOutput.001_metadata.json', 'renderCompositingFileOutput.002_metadata.json', 'renderCompositingFileOutput_metadata.json']
```

moonyuet:
> > It published and rendered successfully with blender but fails in submitting publish job when I tried to use existing rendersetups.
> 
> My gut feeling here is that you have multiple instances writing into the same **folder** using the same filenames perhaps? Anyway, can you pass the metadata json files please? These ones:
> 
> ```
> ['renderCompositingFileOutput.001_metadata.json', 'renderCompositingFileOutput.002_metadata.json', 'renderCompositingFileOutput_metadata.json']
> ```

It could be, is there any validator check on the name of those file output nodes?

moonyuet:
> Please see the publish job submission params in Deadline to see which metadata json it ended up using. :)

I dont see any metadata json as the file output appears to use the relative path as render directory output. See screenshot below.


> 
> Can you also share the scene file + the publish json report from the submission in blender itself?

Like another PR before. I will DM you the scene file.

I enclosed the publish report here, this is super helpful for you to notice the issue.
This contributes to the relative path used in blender 
image
image

[publish-report-250714-22-13.json](https://github.com/user-attachments/files/21217512/publish-report-250714-22-13.json)

BigRoy:
> I enclosed the publish report here, this is super helpful for you to notice the issue.
> This contributes to the relative path used in blender

Thanks for the scene file and the added screenshot but the issue is in the path you've set... a relative path in blender starts with `//` not `/`. For me `/` also writes to `C:/` but I'm worried that there may be chances where `/` would not do that. (It doesn't seem to be using the current working directory at least, e.g. `os.getcwd()`). Anyway, I have considered adding a validator for paths starting with a single slash - the issue is just that those paths may be enitrely valid on Mac/Linux and I wasn't too sure how to really deal with this cross-platform.

Note also that [`os.path.isabs("/blender/test.1001.exr") == True`](https://docs.python.org/3/library/os.path.html#os.path.isabs) because it starts with a slash, so also Python wouldn't be able to detect it not as a relative path (even on Windows). We could do a validation whether we're on Windows, the path starts with `/` and the path does not exist. Something like:
```python
if (
    is_windows
    and path.startswith("/") and not path.startswith("//")
    and not os.path.exists(path)
):
    raise ValueError(
        "Path starts with '/' but is not a valid absolute path. This may render to your "
        "system root and is likely unintended. Set a correct filepath or use two "
        "forward slashes '//' to indicate a relative path from the blender scene file."
    )
```
But it seems extremely specific?

For what it's worth, [here's](https://github.com/ynput/ayon-blender/blob/d7eb2162ef52f26cfa3e2868a788b3cab3dfc252/client/ayon_blender/plugins/publish/collect_render.py#L231-L261) the logic that I'm using to resolve the full path from the path - note specifically the `bpy.path.abspath(path)` call which should return the absolute path (even from e.g. a `//` relative path) as Blender would see it. But apparently it also can't make out the single `/` into anything sensible for the farm?

client/ayon_blender/api/pipeline.py
```diff
@@ -379,9 +381,42 @@ def on_open():
...
+        # TODO: We may want to allow disabling this feature, especially after
+        #  Blender 4.5+ supporting the `{blend_name}` in templates in render
+        #  paths to avoid needing to hardcode the blender scene filename into
+        #  those paths.
+        # Update any render paths if they previously contained the scene name
+        # Warning: if the scene name is `a` before and now `helloworld` then
+        #  this may easily get out of hand by turning `asset` into
+        #  `helloworldsset`, but since filenames tend to be longer and
+        #  contain version numbers, this is not expected to happen often.
+        search_replace_render_paths(blend_name_before, blend_name_after)
```
BigRoy, line 414: @moonyuet @antirotor @philippe-ynput this makes it very convenient - because the user doesn't need to care about manipulating the render paths each time, or having it invalidated on publish to just keep clicking 'repair' each time you publish. However, as the `warning` suggests in the comment there may be edge cases where this may _not_ do what one would want. Thoughts?

moonyuet:
> Thanks for the scene file and the added screenshot but the issue is in the path you've set... a relative path in blender starts with // not /. For me / also writes to C:/ but I'm worried that there may be chances where / would not do that. (It doesn't seem to be using the current working directory at least, e.g. os.getcwd()). Anyway, I have considered adding a validator for paths starting with a single slash - the issue is just that those paths will be entirely valid on Mac/Linux and I wasn't too sure how to really deal with this cross-platform.

This validator can help for this case (the case from my side is some edge case I must admit). I set up all the file output nodes manually and then create render instance. I also make all paths absolute through `File > External Data` too just in case any paths from datablock are not relative(Not quite related to this issue at all).
image

If I set the path to the absolute path, it works fine but there is some issue arose for the naming and frame range being sticked together. 
Instead of `{staging_dir}/{RenderName}.{frame_range}.{image_format}`, it would be `{staging_dir}/{RenderName}{frame_range}.{image_format}`. This is something needed to be fixed so that users can also use their manually tweaked nodes for farm submission. Enclosed with the publish report and screenshot.
[publish-report-250715-15-15.json](https://github.com/user-attachments/files/21227665/publish-report-250715-15-15.json)

image

client/ayon_blender/api/pipeline.py
moonyuet: It's a good idea but I dont think we won't need to include in every blender scene as some of them are only for building content/animation only. Would it be better to be just introduced when there is file ouput node lasted in the scene?

BigRoy:
Thanks @moonyuet - so essentially the request is:

- Validate render output paths endswith `.{frame}.{ext}` enforcing the frame number being prefixed by a `.`?

Is that correct?

client/ayon_blender/api/pipeline.py
moonyuet: > In that case the paths likely wouldn't be set anyway - so there it wouldn't actually be replacing anything... it'd just end up doing nothing. So essentially the behavior is the same as if I would skip it by doing custom checks?

Ahh...My bad, I just read the related function again that you already have skipped for other shader nodes for the searching part. I am perfectly find with this callback.

BigRoy:
> there is something going on when it repairs

Can you elaborate on "something"? What's the issue?

moonyuet:
> We can validate anything that makes sense - aside of checking that frame number separator, anything else?

It should be checking the frame range number separator and nothing else so far.

moonyuet:
> * [ ]  ~Not start with a single slash `/`. If you meant to use a relative path then use `//` at the start of the path.~ Doesn't seem like something we can validate because a single slash is absolutely valid in Mac/Linux, etc. as mentioned [here](https://github.com/ynput/ayon-blender/pull/67#issuecomment-3071157981).

We can leave this as there is further issue arose for windows users. My case is really the edge case only.

moonyuet:
> > For some reason I can't render out GreasePencil and it is totally a blender bug. So we need to give some remarks that the GreasePencil can't be rendered for this case.
> 
> Is this worth making a separate issue for on `ayon-blender` to keep track of? Or?

We can create separate issue to track and we can continue the discussion on that issue

Issue #69, opened 2024-11-07
UNC paths on Windows with forward slashes fail


jrsndl:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Description

Rendering to UNC path with forward slashes colides with Blender notation of current blend file being ```//```, when using ```ayon+settings://blender/RenderSettings/compositing```

### Steps To Reproduce:

1. Set work root to //server/mount

2. Publish Blender render via Deadline

example expected render:
//server/mount/T037_Ayon_test2/shots/blendertest01/work/lgt/renders/blender/blendertest01_lgt_v006

actual render path:
//server/mount/T037_Ayon_test2/shots/blendertest01/work/lgt/**server/mount**/T037_Ayon_test2/shots/blendertest01/work/lgt/renders/blender/blendertest01_lgt_v006

3. Replace all slashes in Compositing - Ayon File output with backslashes and publish again; render location is correct

### Version
Server 1.42, ayon_core 1.06, ayon_blender 0.2.5

### What platform were you running when you found the bug?
Windows

Forcing path to use backslash notation on Windows platform might solve the issue:
https://github.com/ynput/ayon-blender/blob/a3aa297d983447f0cf865e83c5d6d4a5bc129e05/client/ayon_blender/api/render_lib.py#L80C55-L81C1

PR #70, opened 2024-11-07
Render path: Do not enforce backslashes to support UNC path.

BigRoy:
## Changelog Description

Render path: Do not enforce backslashes to support UNC path.

Quick hotfix for #69

@jrsndl could you test this?

## Testing notes:

1. Set up rendering using UNC path
2. Render path should resolve correctly at rendertime.


jrsndl:
> @jrsndl could you test this?
>
Tested and works for me, thank you!

PR #73, opened 2024-11-09
Load Image to comp or shader: fix applying colorspace 

### LINKED ISSUE #72: colorspace is never applied by load_image_compositor and load_image_shader

### Description

This bug is exactly the same on both loaders (because `load_image_shader` is based on `load_image_compositor` ).

Even when a representation has valid colorspace, those loaders don't apply it to the Blender node.

### Steps To Reproduce:

From Blender, open Ayon-->Load
Use it to load a product that has colorspace.
Look at the Blender node : it has ACEScc colorspace, even if the representation is supposed to use another one.

### Additional context:

Here the 1st line's condition is never met, so the 2nd line is never executed :

```
if representation.get("colorspaceData"):
    colorspace: str = representation["colorspaceData"]["colorspace"]
```

The following fix works all right in my tests (in both loaders) :
```
if representation["data"].get("colorspaceData"):
    colorspace: str = representation["data"]["colorspaceData"]["colorspace"]
```

https://github.com/ynput/ayon-blender/blob/b9ea07ec0098a4ca00727e87a6b374f526a13e8e/client/ayon_blender/plugins/load/load_image_compositor.py#L138-L139
https://github.com/ynput/ayon-blender/blob/b9ea07ec0098a4ca00727e87a6b374f526a13e8e/client/ayon_blender/plugins/load/load_image_shader.py#L136-L137

### Relevant log output:

```shell

```

BigRoy:
## Changelog Description

Fix setting of `colorspace`

Fixes #72

## Testing notes:

1. Load image to compositor and load image to shader should correctly set colorspace.

### DIFF
--- a/client/ayon_blender/plugins/load/load_image_compositor.py
+++ b/client/ayon_blender/plugins/load/load_image_compositor.py
@@ -135,8 +135,10 @@ class LoadImageCompositor(plugin.BlenderLoader):
                 image_comp_node.frame_offset = 0
 
         # Set colorspace if representation has colorspace data
-        if representation.get("colorspaceData"):
-            colorspace: str = representation["colorspaceData"]["colorspace"]
+        colorspace_data = representation.get("data", {}).get(
+            "colorspaceData", {})
+        if colorspace_data:
+            colorspace: str = colorspace_data["colorspace"]
             if colorspace:
                 image.colorspace_settings.name = colorspace
 
--- a/client/ayon_blender/plugins/load/load_image_shader.py
+++ b/client/ayon_blender/plugins/load/load_image_shader.py
@@ -133,8 +133,10 @@ class LoadImageShaderEditor(plugin.BlenderLoader):
         image = image_texture_node.image
         representation: dict = context["representation"]
 
-        if representation.get("colorspaceData"):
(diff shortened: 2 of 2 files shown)


LiborBatek (changes requested):
Doesnt work for me...

did try to load ```PNG``` but its set to  ```ACEScc``` instead of ```sRGB```
image

sry dont know how to provide you some log as there is not any.
could I help you somehow with more info to debug?!

LiborBatek:
however its fair to note that for ```EXR``` it reads and sets colorspace well (aka correct ```ACEScg```)

image

BigRoy:
@LiborBatek can you confirm the loaded PNG representation has colorspace data stored on the representation?

Also, we should really start adding the "colorspace" as data into the loader UI so this is much easier to visualize for anyone. Basically implement something similar to this: https://github.com/ynput/OpenPype/pull/5459

LiborBatek:
as far as I can tell (also by inspecting product / version details) there is no colorspace written for my image (been published from ayon PS integration)

Screenshot 2024-11-12 145935

LiborBatek (approved):
Once using ```image``` product with colorspace added (via tray publisher) it seem working correctly and my ```color space``` on file node matches the image one...

image

PR #76, opened 2024-11-18
Chore: Don't check for 'remove' value

iLLiCiTiT:
## Changelog Description
Collect review plugin does not check for `"remove"` value on instance data.

The key is never set. Was deprecated many years ago.

## Testing notes:
1. Nothing should change.
2. Review is rendered and integrated.
3. Review with audio does work as before this PR (don't know if it did work?).


LiborBatek (approved):
Been tested and works ok.

PR #79, opened 2024-11-28
Add project_name as part of the container data

moonyuet:
## Changelog Description
This PR is to add project_name as part of the container data regarding to the implementation of https://github.com/ynput/ayon-core/pull/1005 

Test along with the installed addon with https://github.com/ynput/ayon-core/pull/1005

## Testing notes:
1. Launch Blender
2. Load Assets from any product type(s)

### DIFF
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -576,6 +576,7 @@ def containerise_existing(
         "namespace": namespace or '',
         "loader": str(loader),
         "representation": context["representation"]["id"],
+        "project_name": context["project"]["name"],
     }
 
     metadata_update(container, data)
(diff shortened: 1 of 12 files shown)


BigRoy (approved):
**Didn't test** but code changes look good to me.

LiborBatek (approved):
I was able to load the product and with proper name. Also visible in ```Scene Inventory``` where ```Library project``` been shown too.

image

I guess we are not expecting more than that, right? As it looks correct to me as is now.

BigRoy:
> I guess we are not expecting more than that, right? As it looks correct to me as is now.

You should be able to completely manage it, like updating versions and even switch asset functionality (currently only within the project it was loaded from)

Issue #19, opened 2024-08-09
Extract Review Fails due to max 16 local views opened


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When trying to publish ```Review``` it can fail due to multiple views present in active workfile the publish happens from.

![image](https://github.com/user-attachments/assets/e9cc88da-bee8-4d3b-a1e1-09c1b06f8065)

This seems to happen after mutliple publishes happened in past (15x times plus) from the same workfile...this is just my guess looking on the error shown...

### Expected Behavior:

User have to be able to produce unlimited amount of publishes speaking of any product type including ```Review``` one. Now this not being the case and Ayon fails due to ```Blender``` policies...

We have to ensure / force blender to not produce those unnecessary views or shut those down aka close those viewports.

```
Traceback (most recent call last):
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\publish\extract_playblast.py", line 89, in process
    path = capture(**preset)
           ^^^^^^^^^^^^^^^^^
  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\capture.py", line 87, in capture
    applied_view(window, camera, isolate, options=display_options)
  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\capture.py", line 159, in applied_view
    isolate_objects(window, isolate or objects)
  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\capture.py", line 132, in isolate_objects
    bpy.ops.view3d.localview()
  File "C:\Work\REPO_SW\DCC\BLENDER\4.2.0\WINDOWS\4.2\scripts\modules\bpy\ops.py", line 109, in __call__
    ret = _op_call(self.idname_py(), kw)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Error: No more than 16 local views
```

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Create ```Review``` publish instance\
2. Publish Review 15+ times (versions up to v016)
3. After those it should start failing due to many viewports left opened

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[...]

PR #80, opened 2024-11-28
Return to global view before closing a window

TechnicalDirector:
Blender currently is not clearing the localview when closing a window. (Known bug: https://projects.blender.org/blender/blender/issues/104069) That caused the extract review to fail due to max 16 local views are allowed.

## Changelog Description
I have added a function to restore to the global view before removing the window.

## Testing notes:
1. Create Review publish instance\
2. Publish Review 15+ times (versions up to v016)

### DIFF
--- a/client/ayon_blender/api/capture.py
+++ b/client/ayon_blender/api/capture.py
@@ -101,6 +101,8 @@ def capture(
                 view_context=True
             )
 
+        restore_global_view(window)
+
     return filename
 
 
@@ -117,7 +119,6 @@ ImageSettings = {
     },
 }
 
-
 def isolate_objects(window, objects):
     """Isolate selection"""
     deselect_all()
@@ -133,7 +134,22 @@ def isolate_objects(window, objects):
 
     deselect_all()
 
+def restore_global_view(window, isolate=None):
+    """Exit local view if active.
 
+    Blender currently does not exit localview when closing windows.
+    """
+
+    types = {"MESH", "GPENCIL"}
+    objects = [obj for obj in window.scene.objects if obj.type in types]
+
+    context = create_blender_context(selected=isolate or objects, window=window)
+
+    with bpy.context.temp_override(**context):
+        # Only toggle back if in local view
+        if bpy.context.space_data.local_view:
+            bpy.ops.view3d.localview()
+    
 def _apply_options(entity, options):
     for option, value in options.items():
         if isinstance(value, dict):
--- a/client/ayon_blender/api/capture.py
+++ b/client/ayon_blender/api/capture.py
@@ -101,8 +101,6 @@ def capture(
                 view_context=True
             )
 
-        restore_global_view(window)
-
     return filename
 
 
@@ -134,7 +132,7 @@ def isolate_objects(window, objects):
 
     deselect_all()
 
-def restore_global_view(window, isolate=None):
+def restore_global_view(window):
     """Exit local view if active.
 
     Blender currently does not exit localview when closing windows.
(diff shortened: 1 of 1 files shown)


Dreamink-Official:
That should fix the issue #19

client/ayon_blender/api/capture.py
```diff
@@ -133,7 +134,22 @@ def isolate_objects(window, objects):
 
     deselect_all()
 
+def restore_global_view(window, isolate=None):
```
BigRoy, line 137: Why would we care about the `isolate` behavior? I'd remove that - which simplifies the function quite a bit.

Also, I really have a feeling that somehow this should be part of the independent window context manager - but no idea whether that makes any sense from Blender API point of view.

Dreamink-Official: You are right! It does make more sense to call the function inside independet_window - right before closing it!

LiborBatek (changes requested):
Thx for your contrib, 
I have made some testing in ```blender 4.2.x``` and ```blender 4.3.x``` but still gave me error in the scene which is suffering by the issue of too many views...

```
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.8+dev\ayon_blender\plugins\publish\extract_playblast.py", line 89, in process
    path = capture(**preset)
           ^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.8+dev\ayon_blender\api\capture.py", line 87, in capture
    applied_view(window, camera, isolate, options=display_options)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.8+dev\ayon_blender\api\capture.py", line 173, in applied_view
    isolate_objects(window, isolate or objects)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.8+dev\ayon_blender\api\capture.py", line 131, in isolate_objects
    bpy.ops.view3d.localview()
  File "C:\Work\REPO_SW\DCC\BLENDER\4.2.6\WINDOWS\4.2\scripts\modules\bpy\ops.py", line 109, in __call__
    ret = _op_call(self.idname_py(), kw)
RuntimeError: Error: No more than 16 local views

```

Dreamink-Official:
https://github.com/user-attachments/assets/f4c152eb-a5f4-41ec-95a0-930924e786da

That is strange! Just checked with the newest addon version - added the code and tried it. Still runs perfectly. Would you mind checking again?

We are using blender version: v.4.3.1 RC

LiborBatek:
@Dreamink-Official thank you for your time and evidence! much appreciated...

hmm, maybe I can think of when scene already has that "overflow of viewports" present then this PR does not fix that...aka just when using from the very begining.

Any chance to test on your side too...just find or prepare scene with aformentioned error (present) and try to test publishing from that state...

I will give it a go from fresh new asset/context and see if fixing it (probably will).

I guess if its the case we could probably live with that and approve this PR so it wont show up again anytime! Thanks again!

Cheers!

Dreamink-Official:
You're absolutely right! I forgot to mention that. With the changes I made, I’m just ensuring that the view is properly closed when needed. However, I’m not handling the case you described!

Thanks for pointing that out! Let me know if you need me to adjust anything further. 😊

LiborBatek:
> 
> // Oh and yes. Just checked it on my side too!
> 


Do you think there is possible solution for that occasion too? Would be best to deal with it too...

If not, just say and we shall merge anyway I guess.

Thx

LiborBatek (approved):

moonyuet (approved):

Issue #61, opened 2024-10-15
Removing "BlendScene" publish instance in the Publisher via trash icon removes the actual asset(s)


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Once having the ```blendScene``` publish instance present in the workfile and wanting to remove such instance via dedicated Publisher trash icon actually also removes the assets not just the instance...

![B3d_RemovingPublishInst](https://github.com/user-attachments/assets/48c8b386-8f90-4529-ab3a-8f15dc58e681)

### Expected Behavior:

When using ```Trash``` icon action it should only remove the publish instance and not the actual asset data from the workfile resulting into deleted objects.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Create ```blendScene``` publish instance containing some assets
2. Use ```Remove``` action via trash icon in the Publisher UI
3. Removes all data including assets not just the publish instance

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

PR #83, opened 2024-12-10
Removing instance of BlenderScene won't remove the scene assets.

moonyuet:
## Changelog Description
This PR is to fix the bug of removing BlenderScene instance also deleting the included asset inside it.
Resolve https://github.com/ynput/ayon-blender/issues/61

n/a

## Testing notes:
1. Create BlenderScene Instance with your assets
2. Removing it

### DIFF
--- a/client/ayon_blender/plugins/create/create_blendScene.py
+++ b/client/ayon_blender/plugins/create/create_blendScene.py
@@ -32,3 +32,17 @@ class CreateBlendScene(plugin.BlenderCreator):
                     instance_node.objects.link(data)
 
         return instance_node
+
+    def remove_instances(self, instances):
+
+        for instance in instances:
+            node = instance.transient_data["instance_node"]
+
+            if isinstance(node, bpy.types.Collection):
+                for children in node.children_recursive:
+                    if isinstance(children, bpy.types.Collection):
+                        bpy.data.collections.remove(children)
+                    else:
+                        bpy.data.objects.remove(children)
+
+            self._remove_instance_from_context(instance)


BigRoy (commented):
I can see how and why this works - admittedly though, I feel like this does the exact opposite of what I expect would happen. I would instead expect the objects that are IN the instance to remain existing within the Blender scene and not ONLY live in the Blender Scene instance. It acts solely as an object set / container. Right?

LiborBatek (changes requested):
It partially resolve the issue and do not remove the objects/geometries once deleting publish instance ```blendScene``` but it is not removing also the ```Collection``` used for the publishing as seen on pic below:

Screenshot 2024-12-10 130927

so the instance is removed from the ```Publisher``` UI only and once refreshing it or creating new publish instance ```blendScene``` is leading to duplication and creation of another instance with ```001``` appendix. as seen below, which is undesirable.


Screenshot 2024-12-10 131024


We have to got rig of everything besides the actual scene objects aka user created content.

client/ayon_blender/plugins/create/create_blendScene.py
```diff
@@ -32,3 +32,17 @@ def create(
...
+
+        for instance in instances:
+            node = instance.transient_data["instance_node"]
+
+            if isinstance(node, bpy.types.Collection):
+                for children in node.children_recursive:
+                    if isinstance(children, bpy.types.Collection):
+                        bpy.data.collections.remove(children)
+                    else:
+                        bpy.data.objects.remove(children)
```
BigRoy, line 46: Would just be this? (assuming the object is a collection)
```suggestion
            bpy.data.collections.remove(node)
```

moonyuet: indeed yes

BigRoy (approved):
Didn't test - but code seems fine to me.

LiborBatek (approved):
Now all works as expected! All good...

Issue #75, opened 2024-11-15
Version Up Workfile menu item does no appear


alainxi:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

With the config seen on screenshot ( Settings-->Core-->Tools-->Version Up Workfile-->ON ) , Blender should get a `Version Up Workfile` item in its Ayon menu.
But it does not appear in Blender's menu, because it is not yet implemented.

From Blender's side it's a missing feature.
From Ayon's Web App side it's a description bug, as the hint (see the screenshot below) implies it should work for any DCC.

It was discussed here :   https://community.ynput.io/t/version-up-workfile-does-not-appear-in-blender/2000

![Image](https://github.com/user-attachments/assets/888abdb7-7529-491f-8fe9-513cdeca8c43)

PR #84, opened 2024-12-11
Version up current workfile in ayon menu

moonyuet:
## Changelog Description
This PR is to add version 
Resolve https://github.com/ynput/ayon-blender/issues/75

Should we introduce the validator or integrator to check/save workfile with version?

## Testing notes:
1. Following the testing step in 75 to enable version up workfile
2. Launch Blender and `version up current workfile` would show up in the menu

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -399,7 +399,7 @@ class SetUnitScale(bpy.types.Operator):
 
 
 class VersionUpWorkfile(LaunchQtApp):
-    """Launch Avalon Publisher."""
+    """Perform Incremental Save Workfile."""
 
     bl_idname = "wm.avalon_version_up_workfile"
     bl_label = "Version Up Workfile"
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -397,6 +397,17 @@ class SetUnitScale(bpy.types.Operator):
         return {"FINISHED"}
 
 
+class VersionUpWorkfile(LaunchQtApp):
+    """Launch Avalon Publisher."""
+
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
It works in general (and perform incremental save) ...but there are few things to consider

First the menu item does have wrong tool tip as seen on the pic below:


Screenshot 2024-12-11 101227

second I would follow the maya one by its naming and also put it as a first item in the AYON menu? I quite like that fact

Screenshot 2024-12-11 101134

Lastly (optional) ```Blender``` natively offers ```Incremental Save``` via ```Ctrl+ALT+S``` and it would be cool to use it for AYON up version action too...so it would be as natural as anytime when in blender even without our integration... not sure tho if possible

LiborBatek (changes requested):
Its better on top and with hotkey...which is cool but

image

Add correct name ```Version Up Workfile``` and Correct tooltip as now its off ```Launch Avalon Publisher```... to something like ```perform incremental save``` or so

client/ayon_blender/api/ops.py
```diff
@@ -423,6 +435,24 @@ def draw(self, context):
...
         )
         context_label_item.enabled = False
+        project_name = get_current_project_name()
+        project_settings = get_project_settings(project_name)
+        if project_settings["core"]["tools"]["ayon_menu"].get(
+            "version_up_current_workfile"):
+                layout.separator()
+                layout.operator(
+                    VersionUpWorkfile.bl_idname,
+                    text="Version Up Current Workfile"
```
moonyuet, line 445:
```suggestion
                    text="Version Up Workfile"
```

LiborBatek (approved):
Works like charm now!

image

Issue #88, opened 2025-01-09
Submission to Deadline missing "chunk size" parameter / "Frame per task" not propagated from the publisher UI


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
Once hitting publish "Render" with set ```Frames per task``` to e.g. ```10``` this value is not propagated and submission using default value ```1```

<img width="686" alt="Image" src="https://github.com/user-attachments/assets/fb6e542e-cc32-4167-aa96-c096a1135a15" />

## Expected Behavior

```Frames per task``` if set to non default value must be taken into respect and used when rendering the job on DL.

### Steps To Reproduce:

1. Create ```Render``` publish instance
2. Change value to non default on ```Frames per task```
3. Publish to DL
4. Open DL Monitor and inspect the job and its ```Chunk Size``` value

### Additional context:

Here you can see the comparison between correctly submitted job from ```maya``` host:

<img width="261" alt="Image" src="https://github.com/user-attachments/assets/094d5d41-0094-486c-981e-31c46fc6c8c6" />

and malfunctioning ```blender``` job which is missing that parameter:

![Image](https://github.com/user-attachments/assets/79670993-3963-4118-9f66-681e90816d3d)

### Relevant log output:

```shell

```

iLLiCiTiT:
I think the issue belongs to `ayon-deadline`.

LiborBatek:
could be, but pls note that there were quite a few changes in this area recently (in blender) too

namely this one:

[https://github.com/ynput/ayon-blender/pull/82](url)

BigRoy:
Please confirm whether you're using latest `ayon-deadline`.

LiborBatek:
yes, its been experienced with ```ayon-deadline```  ```0.5.1```

BigRoy:
Great  - can you provide a publish `.json`  report from the submissions? :)

LiborBatek:
Yes, here you have it...note just version is upped by one from ```v024``` to ```v025``` as I had to re-run the publishing for the JSON report otherwise the same scene for sure...

[B3D_publish-report-250109-16-57.json.txt](https://github.com/user-attachments/files/18363990/B3D_publish-report-250109-16-57.json.txt)

BigRoy:
[...]

PR #90, opened 2025-01-11
Make sure the deadline setting correctly propagated to publisher UI

moonyuet:
## Changelog Description
This PR is to make sure the deadline setting correctly propagated to publisher UI and fix the small bug on collecting aov files when the multilayer is on.
Resolve https://github.com/ynput/ayon-blender/issues/88

n/a

## Testing notes:
1. Launch Blender
2. Create Render Instance with mutlilayer options on in ayon setting
3. Publish
4. The deadline parameter should be propagated correctly accordingly to the Publisher UI.

### DIFF
--- a/client/ayon_blender/api/render_lib.py
+++ b/client/ayon_blender/api/render_lib.py
@@ -353,7 +353,8 @@ def set_node_tree(
         # and link it
         for render_layer_node, passes in render_aovs_dict.items():
             render_layer = render_layer_node.layer
-            aov_file_products[render_layer] = []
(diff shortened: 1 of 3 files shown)


LiborBatek (approved):
I have both tested single ```viewLayer``` rendering and multiple ```viewLayers``` and both are working and with propagated ```chunk size``` and also ```priority``` etc.

LGTM

Screenshot 2025-01-13 101848

using ```chunk size``` of ```10``` in my case:

image

PR #94, opened 2025-02-03
Load Material: Add option box to choose material slot to load to

BigRoy:
## Changelog Description

Add option box to choose material slot to load to

image
image
image

Came up here: https://discord.com/channels/517362899170230292/563751989075378201/1335990282944122940

## Testing notes:

1. Allow to choose from the loader which material slot to apply to.
2. This should be tested with multiselection (loading multiple products at once)

### DIFF
--- a/client/ayon_blender/plugins/load/load_image_shader.py
+++ b/client/ayon_blender/plugins/load/load_image_shader.py
@@ -74,12 +74,8 @@ class LoadImageShaderEditor(plugin.BlenderLoader):
         # TODO: We tend to avoid acting on 'user selection' so that the loaders
         #  can run completely automatically, without user interaction or popups
         #  So we may want to investigate different approaches to this.
-        selected_objects = lib.get_selection()
-        for obj in selected_objects:
-            if obj.type in {'MESH', 'SURFACE'}:
-                cur_obj = obj
-                break
-        else:
+        cur_obj = self.get_selected_object()
+        if cur_obj is None:
             self.log.info(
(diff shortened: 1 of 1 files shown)


alainxi:
Thanks for this PR !!!

I tested it in Blender 4.0.2, and selected multiple images at once before using right-mouse-button.

If the object does not already have a material, or if you select `New Material` in the option popup, it creates several materials (one for each image that is loaded).
It would be better to load all selected images into a single new material.

BigRoy:
> I tested it in Blender 4.0.2, and selected multiple images at once before using right-mouse-button.
> 
> If the object does not already have a material, or if you select `New Material` in the option popup, it creates several slots/materials (one for each image that is loaded), which is disturbing. In the case of a new material, I think it would be better to load all selected images into a single material. Because when you load several images, it's usually `normal`, `baseColor`, `Roughness`, etc... which are all intended to connect to the same material and slot.

Of course - valid point and the behavior makes sense code-wise. I'd need to look into how to solve that best without creating a mess ... when I have some spare time.

LiborBatek (approved):
Have been tested in ```blender 4.2.6 LTS``` and ```4.3.x``` and works well for both
without option box and with selection for ```new material```

LGTM

LiborBatek:
@alainxi good points, but these are more quality of life feature requests.... this PR  fixes the error and user should def setup some material so no multiple materials show up if he did not prepare the material first

can be merged as is imho (solves the hard failure atm)

PR #98, opened 2025-02-11
Repair action for fixing render output directory should be correct when enabling multilayer exr

moonyuet:
## Changelog Description
This PR is to fix the repair action of fixing the render output directory.
Fix https://github.com/ynput/ayon-blender/issues/99

n/a

## Testing notes:
1. Launch Blender 
2. Creating Render instance
3. Publish
4. Deleting the instance and recreating
5. Repair the output with validator
6. Publish
7. Should be working when multilayer option is on or off

### DIFF
--- a/client/ayon_blender/plugins/create/create_render.py
+++ b/client/ayon_blender/plugins/create/create_render.py
@@ -43,7 +43,7 @@ class CreateRenderlayer(plugin.BlenderCreator):
         if not bpy.data.filepath:
             version_up_current_workfile()
         else:
-            filepath = version_up(bpy.data.filepath)
+            filepath = bpy.data.filepath
             save_file(filepath, copy=False)
 
         return collection
--- a/client/ayon_blender/plugins/create/create_render.py
+++ b/client/ayon_blender/plugins/create/create_render.py
@@ -1,7 +1,6 @@
 """Create render."""
 import bpy
 
-from ayon_core.lib import version_up
 from ayon_core.pipeline.context_tools import version_up_current_workfile
 from ayon_blender.api import plugin, lib
 from ayon_blender.api.render_lib import prepare_rendering
--- a/client/ayon_blender/plugins/create/create_render.py
+++ b/client/ayon_blender/plugins/create/create_render.py
@@ -1,6 +1,7 @@
 """Create render."""
 import bpy
 
+from ayon_core.lib import version_up
 from ayon_core.pipeline.context_tools import version_up_current_workfile
 from ayon_blender.api import plugin, lib
 from ayon_blender.api.render_lib import prepare_rendering
@@ -42,7 +43,7 @@ class CreateRenderlayer(plugin.BlenderCreator):
         if not bpy.data.filepath:
             version_up_current_workfile()
         else:
-            filepath = bpy.data.filepath
+            filepath = version_up(bpy.data.filepath)
             save_file(filepath, copy=False)
 
         return collection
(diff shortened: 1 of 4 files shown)


client/ayon_blender/plugins/create/create_render.py
```diff
@@ -43,7 +42,7 @@ def create(
         if not bpy.data.filepath:
             version_up_current_workfile()
         else:
-            filepath = version_up(bpy.data.filepath)
+            filepath = bpy.data.filepath
```
BigRoy, line 45: This seems like an odd change?

> It also ensures that creating render instance would version up and save the files when there is existing render output.

Doesn't this do the **opposite**?

BigRoy: But - **why** don't we want that? Aren't we always submitting to generate new renders? why does it need to render over the existing content?

LiborBatek (changes requested):
It works for single layer file formats (exr, png, etc)

Screenshot 2025-02-11 112319

but still not working for ```multilayer``` ```exr``` atm:


Screenshot 2025-02-11 112503

client/ayon_blender/plugins/publish/validate_deadline_publish.py
```diff
@@ -13,6 +14,25 @@
...
+    Returns:
+        str: new_workfile_name
+    """
+    # Use regex to extract the numeric part
+    # Matches one or more digits at the end of the string
+    for directory in old_workfile_dir.split("\\"):
+        old_workfile_match = re.search(r"(\d+)$", directory)
+        if old_workfile_match:
+            old_workfile_name = directory
+            return old_workfile_name
```
BigRoy, line 33: This is overly simplified and would essentially break if there's any other digit in the path apart from the filename.
Also, splitting on backslashes seems VERY Windows-specific?

Also, the docstring mentions `new_workfile_name` which is not an argument to the function.

Similarly, I'm unsure how this implementation "gets the correct workfile directory"? It doesn't seem to check any correctness?

client/ayon_blender/plugins/publish/validate_deadline_publish.py
```diff
@@ -13,6 +14,26 @@
...
+    """
+    # Use regex to extract the numeric part
+    # Matches one or more digits at the end of the string
+    path = Path(old_workfile_dir)
+
+    old_workfile_name = path.parent.name
+    print(old_workfile_name)
+    old_workfile_match = re.search(r"(\d+)$", old_workfile_name)
+    if old_workfile_match:
+        return old_workfile_name
```
BigRoy, line 34: I'm still so confused what this is supposed to be doing?

The type hint is incorrect - since it _only_ returns a string if there's a version match. Next question is, why do we even care about the workfile directory and or version in it?

The logic here seems **very** hardcoded to assuming that the render output ALWAYS renders into an output path where the parent folder is the 'workfile name'?

moonyuet: I believe the original setup is also hardcoded but thanks for reminding as I find even simpler way for the fix. https://github.com/ynput/ayon-blender/blob/114a09a32fc41d15a560ed58cce61ff00797a6a6/client/ayon_blender/api/render_lib.py#L323

LiborBatek (changes requested):
Sry still not work for me when using ```multilayer EXR```

and getting this filepath/filename:

```C:\projects\Blender\Shots\sh010\work\lighting\renders\blender\Blnd_sh010_lighting_v016\ViewLayer\sh010_renderlayerMain_ViewLayer_beauty.####```

LiborBatek (approved):
Screenshot 2025-02-13 111050

Now all works as expected. Also ```Repair``` action produces desired result. Been tested on DL as a render job and output directory been accessible and fine.

Issue #101, opened 2025-02-24
ayon-blender does not load in Blender 4.4


alainxi:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Description

Using the beta version of Blender 4.4 (whose full release is scheduled to appear in 3 weeks), I get an error when starting Blender from the Ayon Launcher (see logs below).

### Steps To Reproduce:

Just launch Blender 4.4 from Ayon's launcher (every Ayon modules are the latest available today).

### Additional context:

Then I replaced  `_global_loaded_modules`  by  `_registered_module_names`  in this line :

https://github.com/ynput/ayon-blender/blob/564e4a48b99282e849a3904757ac25600e35f6e4/client/ayon_blender/api/lib.py#L72

Now the error does not appear anymore.

### Relevant log output:

```shell
Couldn't load user scripts "\\filer1\pub_dev\bin\blender\blenderSetup_4_4"
Traceback (most recent call last):
  File "C:\Users\Alain\AppData\Local\Ynput\AYON\addons\blender_0.2.9\ayon_blender\api\lib.py", line 137, in append_user_scripts
    load_scripts(user_scripts.split(os.pathsep))
  File "C:\Users\Alain\AppData\Local\Ynput\AYON\addons\blender_0.2.9\ayon_blender\api\lib.py", line 89, in load_scripts
    test_register(mod)
  File "C:\Users\Alain\AppData\Local\Ynput\AYON\addons\blender_0.2.9\ayon_blender\api\lib.py", line 72, in test_register
    bpy.utils._global_loaded_modules.append(mod.__name__)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'bpy.utils' has no attribute '_global_loaded_modules'
```

BigRoy:
Thank you - even more reasons to try and see if we can avoid using custom 'install' and maintain that duplicated code from Blender itself.

We should really re-try the `BLENDER_SYSTEM_SCRIPTS ` variable again now it supports multiples in Blender 4.4 as discussed [here](https://discord.com/channels/517362899170230292/1338925629114159115)

alainxi:
I should add that even with my "fix" above, it does not fix the other error that appears when you try to open an Ayon gui from inside Blender (Load, Publish, Workfile, etc...) :

`  Python: RuntimeError: could not create instance of WM_OT_avalon_publisher to call callback function 'execute'`


[...]

PR #102, opened 2025-02-24
Fix operator for Blender 4.4+

BigRoy:
## Changelog Description

Fix support for Blender 4.4+

See: https://developer.blender.org/docs/release_notes/4.4/python_api/

> Python-defined classes based on Blender types (like `Operator`, `PropertyGroup`, etc.) that define their own `__new__`/`__init__` constructors must now call the parent's matching function, and pass on generic positional and keyword arguments.

Fix https://github.com/ynput/ayon-blender/issues/101#issuecomment-2679138882

## Testing notes:

1. Download Blender 4.4 beta
2. Launch with AYON integration

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -200,7 +200,8 @@ class LaunchQtApp(bpy.types.Operator):
     _init_kwargs: Optional[Dict] = dict()
     bl_idname: str = None
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


alainxi:
Thanks for this PR.
I tested it in Blender 4.4 beta, it works all right, the RuntimeError does not happen anymore.
I also tested in Blender 4.0.2, it still works all right.

LiborBatek (approved):
didnt tested it but following others seems ok.

PR #103, opened 2025-02-24
Use `BLENDER_SYSTEM_SCRIPTS` instead of `BLENDER_USER_SCRIPTS` 

BigRoy:
## Changelog Description

Use `BLENDER_SYSTEM_SCRIPTS` instead of `BLENDER_USER_SCRIPTS`  in favor of the improved functionality in Blender 4.4+

Todo:
- [x] Implement logic to detect when it should use this behavior over the original legacy one. (Currently enforces the new behavior which is not ensured to be Blender <=4.3 compatible.

See:
- https://projects.blender.org/blender/blender/issues/127013
- https://projects.blender.org/blender/blender/issues/122512
- https://projects.blender.org/blender/blender/pulls/122689

## Testing notes:

Whenever `AYON_BLENDER_USE_SYSTEM_PATH` is set on Application launch (e.g. by adding that to the blender application's environment settings) it will use the `BLENDER_SYSTEM_PATH` variable now, which will work with Blender 4.4+ and is the recommended approach.

image

So recommended for Blender 4.4+ is:
```json
{
    "AYON_BLENDER_USE_SYSTEM_PATH": "1"
}
```

1. Test with different supported blender versions, and include Blender 4.4 beta in testing.
(4.4+ should definitely work, at least it did for me!)
    - ~~Note that Blender 4.4+ also requires: https://github.com/ynput/ayon-blender/pull/102~~ (now included)
2. Blender should work, tools should work.

### DIFF
--- a/client/ayon_blender/addon.py
+++ b/client/ayon_blender/addon.py
@@ -14,7 +14,7 @@ class BlenderAddon(AYONAddon, IHostAddon):
     def add_implementation_envs(self, env, _app):
         """Modify environments to contain all required for implementation."""
         # Prepare path to implementation script
-        implementation_user_script_path = os.path.join(
+        implementation_script_path = os.path.join(
             BLENDER_ADDON_ROOT,
             "blender_addon"
         )
@@ -26,18 +26,34 @@ class BlenderAddon(AYONAddon, IHostAddon):
             for path in python_path.split(os.pathsep)
             if path
         ]
-        python_path_parts.insert(0, implementation_user_script_path)
+        python_path_parts.insert(0, implementation_script_path)
         env["PYTHONPATH"] = os.pathsep.join(python_path_parts)
 
+        # TODO: What setting or flag will we user before launching Blender
+        #  to define whether it supports the new BLENDER_SYSTEM_SCRIPTS
+        #  with multiple paths?
+        supports_blender_system_scripts = True
+        if supports_blender_system_scripts:
+            self.configure_blender_env(env, implementation_script_path)
+        else:
+            # Old versions of Blender had broken BLENDER_SYSTEM_SCRIPTS
+            # nor supported multiple paths for it. See:
+            # https://projects.blender.org/blender/blender/issues/127013
+            self.configure_blender_pre_44_env(env, implementation_script_path)
+
+        # Define Qt binding if not defined
+        env.pop("QT_PREFERRED_BINDING", None)
+
+    def configure_blender_pre_44_env(
+        self, env: dict, implementation_script_path: str
+    ):
         # Modify Blender user scripts path
         previous_user_scripts = set()
         # Implementation path is added to set for easier paths check inside
         #   loops - will be removed at the end
-        previous_user_scripts.add(implementation_user_script_path)
+        previous_user_scripts.add(implementation_script_path)
 
-        ayon_blender_user_scripts = (
-            env.get("AYON_BLENDER_USER_SCRIPTS") or ""
-        )
+        ayon_blender_user_scripts = env.get("AYON_BLENDER_USER_SCRIPTS") or ""
         for path in ayon_blender_user_scripts.split(os.pathsep):
             if path:
                 previous_user_scripts.add(os.path.normpath(path))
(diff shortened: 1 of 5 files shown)


BigRoy:
@alainxi any chance you could play a bit with this one as well?

alainxi:
- I checked with Blender 4.4 beta, it works all right (Loader, Publish, Workfiles).
Also, our inhouse tools still work (for Blender 4.4 they need BLENDER_USER_SCRIPTS and BLENDER_SYSTEM_EXTENSIONS).


- In Blender 4.0.2, the Ayon menu does not appear anymore with the current state of this PR.  No errors appear anywhere.


- Then I tweaked addon.py to have this :     `supports_blender_system_scripts = False`
It made it work again with Blender 4.0.2 :
Ayon works (Loader, Publish, Workfiles), and our inhouse tools also work (for Blender 4.0.2 they need BLENDER_USER_SCRIPTS).

BigRoy:
@iLLiCiTiT what's a good approach for a 'version' switch behavior per application variant for e.g. for Blender.

I'm wary of parsing the variant string as a "version" because it limits how studios can name their variants - and it feels hacky.
Basically Blender 4.4+ versions should use one way, and anything before Blender 4.4 should use the old. Technically there could be studios with custom compiled versions of Blender where we're not entirely sure where the feature is supported; so 'automatically' detecting it might not be doable either. Plus it needs to affected before launching Blender itself so we can't do a version check inside Blender because we need to adjust env vars to be set.

So a dedicated checkbox OR an env var (to be set in e.g. application variant env) somewhere I think is the only real solution?

client/ayon_blender/addon.py
```diff
@@ -26,18 +26,34 @@ def add_implementation_envs(self, env, _app):
...
+        else:
+            # Old versions of Blender had broken BLENDER_SYSTEM_SCRIPTS
+            # nor supported multiple paths for it. See:
+            # https://projects.blender.org/blender/blender/issues/127013
+            self.configure_blender_pre_44_env(env, implementation_script_path)
+
+        # Define Qt binding if not defined
+        env.pop("QT_PREFERRED_BINDING", None)
+
+    def configure_blender_pre_44_env(
```
iLLiCiTiT, line 47: Both methods should start with `_`.

iLLiCiTiT:
I don't think I'd be helpfull here. Variants that are in my mind:
1. Find out based on available executable path. (e.g. `{executable} --version` and parse it from output)
2. Dedicate env variable that must be set on settings variant to use system path (e.g. `AYON_BLENDER_USE_SYSTEM_PATH=1`).
3. Modify settings model of applications addon for blender. That requires change in applications addon -> blender host still can get the information only using env variable, with current implementation.

alainxi:
If I may :  the  `{executable} --version`  method is the slowest of the 3. It would add another delay, to the already long start that face studios having Blender installed in a shared file-server   (  `\\someServer\blender4_4\blender.exe`  ).

BigRoy:
Thanks - I'll implement 2. for now which I think is the easiest, especially because:
1) would be slow
3) better UX - but close to same result. we could _move_ there from the implementation from 2) too...

BigRoy:
> Dedicate env variable that must be set on settings variant to use system path (e.g. AYON_BLENDER_USE_SYSTEM_PATH=1).

Implemented with https://github.com/ynput/ayon-blender/pull/103/commits/d4bd45ecd08610e15ca4ec2baa982d3d2289f8aa

iLLiCiTiT (approved):
To sum up:
With current changes, it does not change existing behavior, but you can set `"AYON_BLENDER_USE_SYSTEM_PATH"` env variable to `"1"` in applications settings to use system paths.

With that it is possible to use new behavior. If we'll add 4.4 blender variant to applications addon, we can set it by default, so we have "kinda" non-invasive transition time. We can also set `"AYON_BLENDER_USE_SYSTEM_PATH"` to `"0"` for older variants, to be able deprecate it over time.

BigRoy:
@LiborBatek would you like to do one final test run? If so, we could merge this.

PR #105, opened 2025-03-10
Add Link Scene Loader

### LINKED ISSUE #31: Load .blend using "link" instead of "append"

- [X] I have searched the existing issues.

Add support to load files using "link" in blender.

### How would you imagine the implementation of the feature?

Add a loader that links a product.

- [X] I have added the relevant labels to the enhancement request.

_No response_

### Additional context:

Requested on [discord](https://discord.com/channels/517362899170230292/1284996390015074478)

BigRoy:
## Changelog Description

Add a loader to allow linking a **collection** from a `.blend` file. 

### TODO

This is just a quick draft to fix #31 and also fix https://github.com/ynput/ayon-blender/issues/110

- [x] Loading works
- [x] Updating works
- [x] Removing works

This is based off of a stripped version of the code provided by [Submarine](https://submarine.nl/) [here](https://github.com/ynput/ayon-blender/issues/31#issuecomment-2670848081).

## Testing notes:

1. Link any `.blend` file, get expected results
2. Updating linked blend files should keep local overrides.


client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -92,32 +102,37 @@ def process_asset(
 
     def exec_update(self, container: Dict, context: Dict):
         """Update the loaded asset."""
-        collection = container["_collection"]
+        repre = context["representation"]
+        group_name = container["objectName"]
+        collection = bpy.data.collections.get(group_name)
```
BigRoy, line 107: Please avoid getting these by name if we can.

Why can't we use `_collection`?

Preferably we store the explicit object on the container and the actual update/remove logic does not need to "find" the object again - it simplifies the logic, it's faster, safer and more explicit.

moonyuet: The `_collection` can't be imprinted for some reason but there is actually already a node which stores the collection data imprinted inside the container. We can use that one instead for this case.

BigRoy: Sorry, I think you're misinterpreting what I'm saying. I'm not saying to _imprint_ anything differently.

I'm saying that somewhere we're doing `get_containers` which returns the containers. This container can, in the returned data - directly _include_ the actual object we got the imprinted data from so that we don't need to look up the object later by name. The same thing is done in Blender across a few other loaders already...

But looking at it now, that actually [seems to be named `node`](https://github.com/ynput/ayon-blender/blob/c788a1815c039ae8f090de85d0db5d707a9dc819/client/ayon_blender/api/pipeline.py#L605)

It's e.g. used here: https://github.com/ynput/ayon-blender/blob/49727037b3ec42b30487ba4b04be9c2ab2d4609c/client/ayon_blender/plugins/load/load_image_shader.py#L139

Then we don't need to 'look for objects by name' because we directly access the actual Blender python object.

In this case, it seems you should also be able to use `container["node"]` to get the object. Which for you should be the `collection`.

BigRoy (commented):
I'm testing on Blender 4.4 on Windows:

### You can link a file only once

When linking the same file twice, I get:

```

  File "F:\dev\ayon-blender\client\ayon_blender\api\ops.py", line 129, in execute
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "F:\dev\ayon-blender\client\ayon_blender\api\plugin.py", line 495, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "F:\dev\ayon-blender\client\ayon_blender\plugins\load\load_blend_link.py", line 56, in process_asset
    scene.collection.children.link(loaded_collection)

RuntimeError: Error: Collection 'char_hero_blendSceneMain' already in collection 'Scene Collection'
```
Which makes sense and is fine for this first version - however, it does have the "Qt issue" that the error prompt appears BEHIND the AYON Loader UI and is near impossible to get to (escape to kill it really is the only approach it seems?)

**Note:** You can also only link files once if they happen to have the same 'root object names'

### After linking it doesn't show in scene inventory

A linked product does not show in Scene Inventory for me, and hence can't be updated

### The Load Link has option box "Group" enabled by default, but I see no grouping whatsoever?

I don't get any grouping with the `Group` in the option box.
There is no difference when it is disabled?

client/ayon_blender/api/plugin_load.py
```diff
@@ -0,0 +1,216 @@
...
+) -> bpy.types.Collection:
+    """Load a collection to the scene."""
+    with bpy.data.libraries.load(filepath, link=link, relative=False) as (
+        data_from,
+        data_to,
+    ):
+        if lib_container_name is None:
+            lib_container_name = [
+                instance for instance in data_from.collections
+                if instance == "AVALON_INSTANCES"
```
BigRoy, line 110: This seems very wrong. I don't think it's _this_ collection we should enforce to load - it feels like the opposite of we want.
This `AVALON_INSTANCES` is the "loaded content" in a workfile publish (just by luck actually)...

I think we should work towards being more flexible and allowing to load any collection data so it's more universal and less restrictive. Maybe, we can provide some logic to "select" which container to load?

client/ayon_blender/api/plugin_load.py
```diff
@@ -0,0 +1,212 @@
...
+    link=True,
+    lib_container_name = None
+) -> bpy.types.Collection:
+    """Load a collection to the scene."""
+    with bpy.data.libraries.load(filepath, link=link, relative=False) as (
+        data_from,
+        data_to,
+    ):
+        if lib_container_name is None:
+            lib_container_name = data_from.collections[0]
```
BigRoy, line 108: This will fail if the scene has no collections to begin with... which _may_ be possible? Or is that _impossible_ in Blender files I wonder?

Also, this isn't necessarily what I meant. I meant to find the first collection that is NOT a child of any other collections inside the blender file, but maybe the first collection is already that in blender lingo.

I also wonder what happens if on `update` the same named collection does not exist anymore. We may need to test that?

moonyuet: > I also wonder what happens if on update the same named collection does not exist anymore. We may need to test that?

For the updated function, it would still search if there is existing collection and used that one. But the collection has been removed, it would show the corrupted file icon on the linked collection. We might need to re-load the new collection instead if the collection has been deleted.
image

moonyuet:
@BigRoy can you test on the updated commit? I am trying to find the way to relink the collection for the missing libraries, but it seems that there is no precise way to do so. Would be great if you can take a look and give some feedback.

moonyuet:
I did a bit more test and research on the loader, and we can check whether the collection still exists during setting version or upversioning. 
I would just post here for references just in case if we need this in the future when the users don't want the broken link collection during updating or give warning message to users for broken linked collection.

```python
    def _is_collection_existed(self, old_datablock: set, new_datablock: set) -> bool:
        """Check if the currently loaded existed in the updated workfile scene.

        Args:
            old_datablock (set): Asset data from the previous updated workfile
            new_datablock (set): Asset data from the newly updated workfile
        Returns:
            bool: Whether collection still exists in the updated workfile scene.
        """
        existing_objects = [obj for obj in bpy.data.objects if obj in old_datablock]
        if len(new_datablock) == 1 and not existing_objects:
            print(
                "The workfile is not updated as the loaded "
                "collection has been deleted."
            )
            return True
        return False

```

LiborBatek:
I was also able to version manage the rig even if changing the name of the collection before publish e.g. former name ```Character``` been changed to ```Char```

and it still worked ok which is good!

The issues arrises when bringing more than one rig asset of the same kind (asset and version) and not having any ```Library Override``` turned on as blender for some weird reason not alterning the name marked by red on the scrn below


Screenshot 2025-06-17 145821

then it causes the already reported traceback:

```

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\ops.py", line 129, in execute
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\plugin.py", line 495, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\plugins\load\load_blend_link.py", line 72, in process_asset
    scene.collection.children.link(loaded_collection)

RuntimeError: Error: Collection 'soldier_blendSceneRig' already in collection 'Scene Collection'
```

I guess we should enforce the ```Make library override``` to not make this happen (or at least thats what I can think of atm)

LiborBatek:
So Im able to ```Link``` the ```rig``` product now without any issues which is good!

however I cant get more instances linked just one. When trying to load another instance Im getting error:

```

             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\plugin.py", line 495, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_blend_link.py", line 65, in process_asset
    loaded_collection = load_collection(
                        ^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\plugin_load.py", line 137, in load_collection
    asset_container.objects.link(obj)

RuntimeError: Error: Object 'vanguard_visor' already in collection 'vanguard_01_rigMain'
```

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,152 @@
...
+            namespace=namespace or "",
+            nodes=[loaded_collection],
+            context=context,
+            loader=self.__class__.__name__,
+        )
+        # TODO: Store loader options for later use (e.g. on update)
+        # Store the loader options on the container for later use if needed.
+        extra_data =  {
+            "options": options,
+            "lib_name": os.path.basename(filepath)
```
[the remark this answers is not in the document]
BigRoy: Sorry - I don't get it?

- What's the value of `library.name` and `library.name_full` when we load it?
- Does that value persist on scene reopen?
- If so, we can store that value onto the container, so that `container["lib_name"] = library.name_full` (or `library.name` if that's sufficiently unique in blender)
- Then the `get_library_by_name` would just do `library.name_full == container["lib_name"]`.

No filename is involved at that point, and we're taking whatever blender reports as the library name... which I assume is 'consistent'?

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,173 @@
...
+            elif child.override_library and child.override_library.reference:
+                return child.override_library.reference.library
+
+        return None
+
+    def _get_library_by_name(self, container: Dict) -> bpy.types.Library:
+        """Get the library by filename."""
+        lib_name = os.path.basename(container["libpath"])
+        for library in bpy.data.libraries:
+            print(library.name_full)
```
BigRoy, line 171:
```suggestion
```

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,173 @@
...
+            "loader": str(self.__class__.__name__),
+            "representation": context["representation"]["id"],
+            "libpath": filepath,
+            "parent": context["representation"]["versionId"],
+            "productType": context["product"]["productType"],
+            "objectName": group_name,
+            "loaded_collection": loaded_collection,
+            "project_name": context["project"]["name"],
+            "options": options,
+        }
```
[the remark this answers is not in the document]
moonyuet: I want to align with how the data stored with other loaders in blender, the options key is from the contributor's implementation. loaded_collection is useful for the updating function(I am working in progress to fix it now).

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,173 @@
...
+                return child.library
+            # With override library
+            elif child.override_library and child.override_library.reference:
+                return child.override_library.reference.library
+
+        return None
+
+    def _get_library_by_name(self, container: Dict) -> bpy.types.Library:
+        """Get the library by filename."""
+        lib_name = os.path.basename(container["libpath"])
```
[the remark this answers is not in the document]
moonyuet: This function would be removed eventually after the change of updating function

LiborBatek (changes requested):
Even tho I can now bring multiple  entities of the same asset Im not able to version manage those atm.

The same goes for single asset only...not able to verion manage it.



Also no any traceback inprintent whatsoever...

Screenshot 2025-06-23 130442

Note that all three assets should be set to ```ver011```  aka the yellowish colored one.

moonyuet:
@LiborBatek I am afraid that link scene loader updating only works for one asset only and if we need to update multiple, all the assets would still be updated to the latest library path we have asked blender to use. 
You can take reference from this video, I am not sure if it is the same case in your side.

https://github.com/user-attachments/assets/596fc5fa-9390-47ab-a33c-9b3bb4fcb99b

alainxi:
If library.reload() does not work :

- There is a dirty trick, using the `relocate` GUI method :  bpy.ops.outliner.lib_operation(context_override, type='RELOCATE')
https://docs.blender.org/api/current/bpy.ops.outliner.html#bpy.ops.outliner.lib_operation
It's probably too hacky.

- Instead, I remember seeing people saving the scene, and loading it, which forces Blender to re-link everything. Sometimes still needing a second call of library.reload()
Sorry I don't find my source anymore, it was more than a year ago.

BigRoy:
I can really recommend joining the Blender dev chat channels for stuff like this - you'll get amazing responses from very tech savvy people of whom most are actual blender devs. See [Blender Chat](https://chat.blender.org/).

The Pipeline & I/O channel is pretty good. :)

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,209 @@
...
+        for library in bpy.data.libraries:
+            if lib_name == library.name_full:
+                return library
+
+    def _update_library_path_by_node(self, collection:bpy.types.Collection, libpath:str):
+        for node in collection.children:
+            if node and node.library:
+                node.library.name = os.path.basename(libpath)
+                node.library.filepath = libpath
+                node.library.reload()
```
BigRoy, line 209: Why would `node` not be valid? :P Is the `if node`.

Also, wouldn't ALL the nodes of the loaded collection essentially be using the same library object? If so, you're now reloading the library TONS of times, which will likely be slow?

Should we instead be doing a first loop where we get all the libraries (in a unique manner so we don't have duplicates). And then just do:
```python
for library in libraries:
    library.name = os.path.basename(libpath)
    library.filepath = libpath
    library.reload()
```

client/ayon_blender/api/lib.py
```diff
@@ -164,6 +164,30 @@ def set_app_templates_path():
...
+    Returns:
+        The name of the container assembly
+    """
+    node_name = f"{context['folder']['name']}_{name}"
+    if namespace:
+        node_name = f"{namespace}:{node_name}"
+    if suffix:
+        node_name = f"{node_name}_{suffix}"
+
+    return node_name
```
BigRoy, line 188: I think this function name is confusing, should this be `create_container_name` or `define_container_name`. `get_*` sounds like it's getting it from an existing instance in the scene?

BigRoy: Also, I think this should just be a function inside `pipeline.py` since it's very pipeline related logic and not just 'blender libs'

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,209 @@
...
+        return len(match_count) > 1
+
+    def _get_or_build_library_by_path(self, libpath: str) -> bpy.types.Library:
+        """Get the library by filepath. If there is no any
+        associated library to the path, the related library is loaded
+        accordingly."""
+        lib_name = os.path.basename(libpath)
+        for library in bpy.data.libraries:
+            if lib_name == library.name_full:
+                return library
```
BigRoy, line 188: Why aren't we just comparing by the [library `filepath`](https://docs.blender.org/api/current/bpy.types.Library.html#bpy.types.Library.filepath)?


So that we just check - is this file already loaded? Yes, use it... otherwise load new one.

client/ayon_blender/plugins/load/load_blend_link.py
moonyuet: It works to load multiple libraries but for some reasons it doesn't update things we want to. But we can try to test with the commit again.

BigRoy: If certain code is convoluted or odd with the reason because "the normal way is buggy" please add a comment to the code describing why the odd workaround is needed.

moonyuet:
@LiborBatek can you test on this again? 
The updating should be working for single assets only but it doesn't work due to the issue found on reloading library datablock in blender(It does not act similar to what maya references does).
I also enclosed the video here.
https://github.com/user-attachments/assets/5beac00b-be45-4618-b5ec-289b6e1ffa9e

LiborBatek:
thx for the vid but you have no library overrides active....this works for me too, but once applied Library override on the Armature then it wont update at all...

Still not working as supposed to, sry.  Here you can cleary see it...works till overrides present...the model v12 (purple) vs v11 (yellowish)....dont switch once override active

https://github.com/user-attachments/assets/7d5f8b07-507d-4d1b-b4a3-cfec3f384267

LiborBatek:
now it successfully switches version via ```Manage>Set Version``` with overrides on place however if any animation been made it will loose it after updating...this also used to worked back then (older commits).

I will need to investigate it more, as I have not performed any save action before managing my version (not saving my active workfile which could maybe the source of the issue tho).

moonyuet:
@LiborBatek it should be updating with/without shader and animation if you update the latest commit.

LiborBatek:
> @LiborBatek it should be updating with/without shader and animation if you update the latest commit.

hmm still just swaps the version but the animation is gone....pretty much still the same behaviour sry.

Also no any errors shown.

moonyuet:
> Stupid question - but we seem to be doing tons of stuff to findf the right `library`, etc. from a 'collection' container?
> 
> Can't we just imprint the data on the library datablock? (is that possible?) so that the actual library itself becomes the AYON container? Which means that technically we also even wouldn't need the "collection"?
> 
> Seems like putting the imprinted data on the object we're actually managing here makes it the easiest approach?
> 

We actually try to do get the library from `container["node"]` or get the library from the children in `container["node"]`, the former one is updated with the library path but it doesn't reload to the children. The latter one works for updating the linked texture when linking but does not work anymore when users manually overriding it. 

The current updating function can update the linked asset successfully for both linking and overriding approach by the users.

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,195 @@
...
+            or self._get_library_by_prev_libpath(container)
+        )
+        filepath = self.filepath_from_context(context)
+        new_filename = os.path.basename(filepath)
+
+        # Update library filepath and reload it if there is library
+        if library:
+            library.name = new_filename
+            library.filepath = filepath
+            library.reload()
```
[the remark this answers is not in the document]
moonyuet: What if we just update the library filepath? When same library name loaded it will be renamed to the libraryname.001 and so on.
```suggestion
        # Update library filepath and reload it if there is library
        if library:
            library.filepath = filepath
            library.reload()
```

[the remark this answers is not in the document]
BigRoy: It still wouldn't solve that the other `container` would still have the wrong `representation` value imprinted.

client/ayon_blender/plugins/load/load_blend_link.py
```diff
@@ -0,0 +1,255 @@
...
+            default=False,
+            tooltip="Add a library override for the loaded asset.",
+        ),
+        BoolDef(
+            "use_arbitrary_loc_for_library",
+            label="Use Arbitrary Location for Library",
+            default=False,
+            tooltip="Store the library in arbitrary location"
+                    "Use with caution when blend file is big.",
+        )
```
BigRoy, line 46: Sorry, but what's this for? This seems odd.

moonyuet: I admit it is odd and extremely hacky. It ensures updating the multiple assets correctly, it may be disturbing but it can temporarily resolve the issue happened in https://github.com/ynput/ayon-blender/pull/105#issuecomment-2996665661 
It may not be pretty solution for sure (we need extra callback for this to check on the linked scene objects if saved for example), but we can still get multiple assets updated without collapsing their loaded libraries when all the linked objects are all same versions (with the same library). Blender would remove unused libraries when they are linked everywhere.

BigRoy: I think this is the wrong fix - we don't want to be copying files over.. this won't be trivial to manage, especially in multi-OS environment, etc. I don't think this is a reasonable workaround in any way. @philippe-ynput thoughts?

LiborBatek (changes requested):
Generally it works now when single product linked as library and  can be version managed too without loosing any animation (need to be set as local data block first- I was missing this workflow blender specific step).

When having multiple same products linked in - version manage is not distinguishing among its versions/ library file paths... as expected and not working and replaces all of the linked products at once (blender limitation I guess, but sry if not thats just my assumption from  conversation taking place in here)

Once using the new option when trying to link asset:


Screenshot 2025-07-07 155928


Im getting errors when trying version manage:

```

             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_blend_link.py", line 136, in exec_update
    self.update_library_filepath(

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_blend_link.py", line 251, in update_library_filepath
    shutil.copy(filepath, updated_libpath)

  File "C:\Work\REPO_SW\DCC\BLENDER\4.5.0\WINDOWS\4.5\python\lib\shutil.py", line 431, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)

  File "C:\Work\REPO_SW\DCC\BLENDER\4.5.0\WINDOWS\4.5\python\lib\shutil.py", line 258, in copyfile
    with open(dst, 'wb') as fdst:
         ^^^^^^^^^^^^^^^

FileNotFoundError: [Errno 2] No such file or directory: '//.linked_folder\\soldier_01_blendSceneRig\\Blnd_soldier_blendSceneRig_v014.blend'

```

also worth noting that when both options enabled it does not create any override too... ignoring the knob state "enabled"


and here are the files located in the working folder (seems correct)

image

LiborBatek (approved):
So now all works as we have decided it should be working. To summarize it:

User can Link product just once, if trying again Instance Collection happens instead. This way we prevent to load the same Library more than one time.

Screenshot 2025-07-10 113853


All Manage functions been fully functional and can be triggered by user.

Havent spotted any issues so I guess approved.

Note: Currently we are limiting user on using the same asset (namely speaking of ```Rig```)  just single time which is not on par with native Blender workflow as it is possible to link the same library multiple times and make an anim override on each separately. However due to the fact we cant exclusively version manage each instance of the same asset we do not allow it atm. There happens to be also present blender issue regarding overrides on linked libraries too so it makes sense to postpone such capability of loading multiple times for later.

PR #107, opened 2025-03-20
Raise descriptive error when Blender launched without -python-use-system-env argument

BigRoy:
## Changelog Description

Raise a more descriptive error when a user has launched AYON Blender without the `--python-use-system-env` argument

It would now error something like:
```
  File "<stdin>", line 2, in <module>
ModuleNotFoundError: No module named 'ayon_core'

The above exception was the direct cause of the following exception:

  File "<stdin>", line 5, in <module>
ModuleNotFoundError: No module named 'ayon_core'. This usually happens if Blender is launched without the --python-use-system-env argument. Make sure you have the Blender application 'Arguments' configured correctly so that Blender has access to the launched context's PYTHONPATH.
```

## Testing notes:

1. Remove the `--python-use-system-env` argument from application settings
2. Launch Blender
3. The raised error should now hint at what's most likely wrong.

### DIFF
--- a/client/ayon_blender/blender_addon/startup/init.py
+++ b/client/ayon_blender/blender_addon/startup/init.py
@@ -1,6 +1,13 @@
-from ayon_core.pipeline import install_host
-from ayon_blender.api import BlenderHost
-
+try:
+    from ayon_core.pipeline import install_host
+    from ayon_blender.api import BlenderHost
+except ModuleNotFoundError as exc:
+    raise ModuleNotFoundError(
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Correctly gives the detailed info when PYTHONPATH argument been missing
```
ModuleNotFoundError: No module named 'ayon_core'. This usually happens if Blender is launched without the --python-use-system-env argument. Make sure you have the Blender application 'Arguments' configured correctly so that Blender has access to the launched context's PYTHONPATH.
```

All good!

PR #104, opened 2025-03-03
Extract FBX or layout: Avoid IndexError if object name does not contain `:`

BigRoy:
## Changelog Description

Extract FBX or layout: Avoid IndexError if object name does not contain `:`

## Additional review information

Fix https://github.com/ynput/ayon-blender/issues/45

## Testing notes:

This may only work with Blender versions 4.1 and lower, see: https://github.com/ynput/ayon-blender/issues/45#issuecomment-2694628505 and may need _additional_ fixes for Blender 4.1+

1. Load some rig and animate it
2. Create Animation publish instance
3. Enable FBX optional flag for extracting FBX too
4. Publish should succeed

Issue #45, opened 2024-09-18
Extract Animation (FBX) not working and fails with error


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When having ```Animation``` publish instance type present and want to publish ```FBX``` repre (optional) it fails and gives following error:

```
DEBUG: Performing extraction..
Traceback (most recent call last):
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.3\ayon_blender\plugins\publish\extract_fbx_animation.py", line 125, in process
    original_name = armature_name.split(':')[1]
                    ~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

### Expected Behavior:

Animation product should also work not only with Alembic but also with ```FBX``` repre type.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1.Load some rig and animate it
2.Create ```Animation``` publish instance
3. Enable ```FBX``` optional flag for extracting FBX too
4. Publish will fail due to the error

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

sosoyan:
I bumped into this recently and pushed a quick fix here, can PR if it helps.
https://github.com/sosoyan/ayon-blender/commit/27a9b310533c2a66b6e24bc034d0e1fe25760ff7

PR #111, opened 2025-04-25
Extract Fbx Animation: `bake_action_objects` support Blender 4.1+

moonyuet:
## Changelog Description
This PR is to fix the Attribute Error shown with `bake_action_objects`  due to the changes of the argument `bakeOptions`

Related to https://github.com/ynput/ayon-blender/pull/104

## Testing notes:
1. Launch Blender 4.0 and 3.X.X 
2. Create Animation and publish
3. Both should be published successfully

### DIFF
--- a/client/ayon_blender/plugins/publish/extract_fbx_animation.py
+++ b/client/ayon_blender/plugins/publish/extract_fbx_animation.py
@@ -8,6 +8,7 @@ import bpy_extras.anim_utils
 from ayon_core.pipeline import publish
 from ayon_blender.api import plugin
 from ayon_blender.api.pipeline import AVALON_PROPERTY
+from ayon_blender.api.lib import get_blender_version
 
 
 def get_all_parents(obj):
@@ -132,13 +133,34 @@ class ExtractAnimationFBX(
         max_frame = min(starting_frames)
         min_frame = max(ending_frames)
 
-        # We bake the copy of the current action for each object
-        bpy_extras.anim_utils.bake_action_objects(
-            object_action_pairs,
-            frames=range(int(min_frame), int(max_frame)),
-            do_object=False,
-            do_clean=False
-        )
+        if get_blender_version[0] >= 4:
+            # We bake the copy of the current action for each object
+            bake_options = bpy_extras.anim_utils.BakeOptions(
+                only_selected=True,
+                do_pose=True,
+                do_object=False,
+                do_visual_keying=True,
+                do_constraint_clear=True,
+                do_parents_clear=True,
+                do_clean=False,
+                do_location=True,
+                do_rotation=True,
+                do_scale=True,
+                do_bbone=True,
+                do_custom_props=True
+            )
+            bpy_extras.anim_utils.bake_action_objects(
+                object_action_pairs,
+                frames=range(int(min_frame), int(max_frame)),
+                bake_options=bake_options
+            )
+        else:
+            bpy_extras.anim_utils.bake_action_objects(
+                object_action_pairs,
+                frames=range(int(min_frame), int(max_frame)),
+                do_object=False,
+                do_clean=False
+            )
 
         for obj in bpy.data.objects:
             obj.select_set(False)
--- a/client/ayon_blender/plugins/publish/extract_fbx_animation.py
+++ b/client/ayon_blender/plugins/publish/extract_fbx_animation.py
@@ -133,8 +133,8 @@ class ExtractAnimationFBX(
         max_frame = min(starting_frames)
         min_frame = max(ending_frames)
 
-        ver_major, ver_min, _ = get_blender_version()
-        if ver_major >= 4 and ver_min >= 1:
+        blender_version = get_blender_version()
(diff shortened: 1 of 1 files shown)


client/ayon_blender/plugins/publish/extract_fbx_animation.py
```diff
@@ -132,13 +133,34 @@ def process(self, instance):
...
         min_frame = max(ending_frames)
 
-        # We bake the copy of the current action for each object
-        bpy_extras.anim_utils.bake_action_objects(
-            object_action_pairs,
-            frames=range(int(min_frame), int(max_frame)),
-            do_object=False,
-            do_clean=False
-        )
+        if get_blender_version[0] >= 4:
```
BigRoy, line 136: Didn't they change it in version 4.1 instead of 4.0?

See here: https://github.com/ynput/ayon-blender/issues/45#issuecomment-2694696220

BigRoy: I see what you mean, but that whole point is worded in an odd way:

> The parameters of anim_utils.bake_action() have been combined into a [dataclass](https://docs.python.org/3/library/dataclasses.html) anim_utils.BakeOptions ([5024c0ef97](https://projects.blender.org/blender/blender/commit/5024c0ef97aa3441d6f4774515d1103ea038d561)). Contrary to the default values of the parameters in Blender 4.0, where some were True and others were False, in this dataclass there are no defaults, and thus all values need to be specified. This helps in local understandability, as there is no need to know which options are there but left out, so using their defaults.

It basically mentions that `4.0` is using the "old" way. The point itself isn't that weirdly phrased - it's just odd that it appears in Blender 4.0 notes... :/ 

This is the API documentation for Blender 4.0 which does NOT seem to use the new way yet: https://docs.blender.org/api/4.0/bpy_extras.anim_utils.html#bpy_extras.anim_utils.bake_action_objects

client/ayon_blender/plugins/publish/extract_fbx_animation.py
```diff
@@ -132,13 +133,35 @@ def process(self, instance):
...
-            object_action_pairs,
-            frames=range(int(min_frame), int(max_frame)),
-            do_object=False,
-            do_clean=False
-        )
+        ver_major, _, _ = get_blender_version()
+        if ver_major >= 4:
+            # We bake the copy of the current action for each object
+            bake_options = bpy_extras.anim_utils.BakeOptions(
+                only_selected=True,
```
BigRoy, line 140: The old way also has the `only_selected` argument it seems, which defaults to `False` and wasn't used.... was that actually a bug and should the old way also specify `only_selected=True`? or should the new way also be using `False`?

It seems we're not making any selections - as such, `only_selected=True` seems wrong?

```suggestion
                only_selected=False,
```

client/ayon_blender/plugins/publish/extract_fbx_animation.py
```diff
@@ -132,13 +133,35 @@ def process(self, instance):
...
+                do_pose=True,
+                do_object=False,
+                do_visual_keying=True,
+                do_constraint_clear=True,
+                do_parents_clear=True,
```
BigRoy, line 145: In the old way the defaults for these were `False` and we were not passing them. As such, this seems to be 'new' behavior if we would enable them now. Should these remain disabled?

```suggestion
                do_constraint_clear=False,
                do_parents_clear=False,
```

client/ayon_blender/plugins/publish/extract_fbx_animation.py
moonyuet: I tested it just now and you are right about that. I updated the changes already.

client/ayon_blender/plugins/publish/extract_fbx_animation.py
```diff
@@ -132,13 +133,35 @@ def process(self, instance):
...
 
-        # We bake the copy of the current action for each object
-        bpy_extras.anim_utils.bake_action_objects(
-            object_action_pairs,
-            frames=range(int(min_frame), int(max_frame)),
-            do_object=False,
-            do_clean=False
-        )
+        ver_major, ver_min, _ = get_blender_version()
+        if ver_major >= 4 and ver_min >= 1:
```
BigRoy, line 137: This unfortunately wouldn't work in Blender 5.0 would be released because it wouldn't pass the minor check.
```suggestion
        blender_version = get_blender_version()
        if blender_version >= (4, 1, 0):
```

E.g.

```python
>>> (4, 1, 0) >= (4, 0, 0)
True
>>> (4, 0, 0) >= (4, 1, 0)
False
>>> (3, 9, 2) >= (4, 1, 0)
False
>>> (4, 2, 0) >= (4, 1, 0)
True
>>> (5, 0, 0) >= (4, 1, 0)
True
```

BigRoy (approved):
**Except** for [the version check](https://github.com/ynput/ayon-blender/pull/111/files#r2060255467), this looks good to me.

Issue #113, opened 2025-05-05
Strip extra namespaces on import / publish and avoid truncation


philippe-ynput:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

from client:
> I'm noticing  that there is still the extra namespace issue where objects keep the cascade namespace for example `05_pipeline_01_blendSceneLookdev:05_pipeline_01_modelMain:cube_` and the names gets truncated.

### Suggested implementation?

from @moonyuet:
> we can do some enhancement to stripe the namespace of loaded instance(in your case: 05_pipeline_01_modelMain) and remove the loaded container metadata during publishing. And once you publish the data with blenderScene(or any product type), it won't load the previous container metadata with the striped namespace

### Describe alternatives you've considered:

None

BigRoy:
I am admittedly still confused _why_ we must strip the namespaces? If there is an issue with the length of the object names I wonder if it was ever a good idea to actually starting adding 'fake' namespaces... :')

alainxi:
In Blender 5.0 (scheduled to be released in november 2025), the maximum data-block names length will be increased from 63 to 255 bytes.
This will allow using names of roughly up to 250 characters.

https://developer.blender.org/docs/release_notes/compatibility/ 

I also logged it here :   https://github.com/ynput/ayon-blender/issues/118

PR #114, opened 2025-05-07
Publishing BlenderScene product should strip namespace and container data 

moonyuet:
## Changelog Description
This PR is to make sure publishing blenderScene product will strip namespace and container data
Resolve https://github.com/ynput/ayon-blender/issues/113

Make sure testing the scenarios below:
1. Publish with loaded asset
2. Publish without loaded asset
3. Publish with or without loaded asset

## Testing notes:
1. Launch Blender 
4. Create BlenderScene product with/without the loaded assets
5. Publish
6. Create New Scene in Blender
7. Load the published asset

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -617,7 +617,12 @@ def strip_namespace(containers):
     ]
     original_namespaces = {}
     for node in nodes:
-        for child in node.children_recursive:
+        if isinstance(node, bpy.types.Collection):
+            children = node.children_recursive
+        elif isinstance(node, bpy.types.Object):
+            children = node.children
+
+        for child in children:
             original_name = child.name
             if ":" not in original_name:
                 continue
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -621,14 +621,14 @@ def strip_namespace():
             original_name = children.name
             namespace, name = original_name.rsplit(':', 1)
             children.name = name
-            namespace_dict[node] = namespace
+            namespace_dict[children] = namespace
 
     try:
         yield
 
     finally:
         for node in nodes:
-            namespace = namespace_dict[node]
             for children in node.children_recursive:
+                namespace = namespace_dict[children]
                 name = children.name
                 children.name = f"{namespace}:{name}"
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -590,11 +590,11 @@ def get_blender_version():
 
 
 @contextlib.contextmanager
-def strip_container_data():
+def strip_container_data(containers):
     """Remove container data during context
     """
     container_data = {}
-    for container in pipeline.ls():
+    for container in containers:
         node = container["node"]
         container_data[node] = dict(
             node.get(pipeline.AVALON_PROPERTY)
@@ -609,11 +609,11 @@ def strip_container_data():
 
 
 @contextlib.contextmanager
-def strip_namespace():
+def strip_namespace(containers):
     """Strip namespace during context
     """
     nodes = [
-        container["node"] for container in pipeline.ls()
+        container["node"] for container in containers
     ]
     namespace_dict = {}
     for node in nodes:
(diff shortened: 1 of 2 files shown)


client/ayon_blender/api/lib.py
```diff
@@ -587,3 +587,48 @@ def get_blender_version():
...
+    nodes = [
+        container["node"] for container in pipeline.ls()
+    ]
+    namespace_dict = {}
+    for node in nodes:
+        for children in node.children_recursive:
+            original_name = children.name
+            namespace, name = original_name.rsplit(':', 1)
+            children.name = name
+            namespace_dict[node] = namespace
```
BigRoy, line 624: Shouldn't this be assigning it to `children` as key?

Also, instead of `children` the variable should be `child` since it's a single entry.

moonyuet: Fix in [3b3602d](https://github.com/ynput/ayon-blender/pull/114/commits/3b3602d0d48b33cb40f95f8cab231b68929674f4)

LiborBatek:
I have tested it and tried to publish ```blendScene``` product but I cant see any difference in namespaces when using ordinary ```model``` product types inside of such product (scene).

so my question is...
so it should only count for when using ```blendScene``` products loaded and publishing the ```blendScene``` product from such scene containing other ```blendScene``` products ? as other types such ```model``` products not affected by any means speaking of stripping namespaces...

or at least thats my initial assumption. any insights welcomed @moonyuet

moonyuet:
> so my question is... so it should only count for when using `blendScene` products loaded and publishing the `blendScene` product from such scene containing other `blendScene` products ? as other types such `model` products not affected by any means speaking of stripping namespaces...


> so my question is... so it should only count for when using `blendScene` products loaded and publishing the `blendScene` product from such scene containing other `blendScene` products ?
Yes. It would only strips the namespace and container data  for the loaded asset when publishing and loading blendScene, but it won't strip the namespace and container data when publishing and loading other products like model, rig etc.

philippe-ynput:
@LiborBatek can you please do a final test when you have a moment so we may close this one ASAP ?
Many thanks ! :-)

BigRoy (commented):
I wonder if it makes slightly more sense to have the context manager take the nodes as inputs, so that you wouldn't need to query `pipeline.ls()` twice, just because it could be a decent speed improvement for very large scenes since it wouldn't need to iterate all materials, all compositing nodes and all scene collections twice.

It'd purely be an optimization though.

moonyuet:
@LiborBatek can you give one more test?
@BigRoy can you take a look at the code again?

client/ayon_blender/api/lib.py
```diff
@@ -587,3 +587,48 @@ def get_blender_version():
...
+
+    try:
+        yield
+
+    finally:
+        for node in nodes:
+            for children in node.children_recursive:
+                namespace = namespace_dict[children]
+                name = children.name
+                children.name = f"{namespace}:{name}"
```
BigRoy, line 634: 1. This would fail if a node has no `:` in the name.
2. Finally iteration can be simplified to go over the `namespace_dict`
3. `children` variable is confusing, because it refers to a single node. Should be `child`.

```suggestion
    original_namespaces = {}
    for node in nodes:
        for child in node.children_recursive:
            original_name = child.name
            if ":" not in original_name:
                continue
            namespace, name = original_name.rsplit(':', 1)
            child.name = name
            original_namespaces[child] = namespace

    try:
        yield
    finally:
        for node, original_namespace in original_namespaces.items():
           node.name = f"{namespace}:{name}"
```

BigRoy (approved):
Didn't test - but code LGTM.

moonyuet:
I have tested in my side. Looks good.

https://github.com/user-attachments/assets/914d3b6a-0519-4628-a9ab-8a966807db5a

client/ayon_blender/api/lib.py
```diff
@@ -587,3 +587,53 @@ def get_blender_version():
...
+                continue
+            namespace, name = original_name.rsplit(':', 1)
+            child.name = name
+            original_namespaces[child] = namespace
+
+    try:
+        yield
+    finally:
+        for node, original_namespace in original_namespaces.items():
+           node.name = f"{original_namespace}:{name}"
```
BigRoy, line 639: Indentation mismatch?
```suggestion
            node.name = f"{original_namespace}:{name}"
```

Issue #30, opened 2024-09-11
Remove OpenPype and avalon references


iLLiCiTiT:
## Description
Make `AYON` and ayon naming as first class citizen for all functionality and metadata naming in repository.

Ideal goal is that `openpype`, `pype` and `avalon` are not mentioned anywhere in codebase. But that probably won't be possible because of backwards compatibility.

## Codebase requirements
- [ ] Instances and containers do use `AYON_INSTANCE_ID` and `AYON_CONTAINER_ID` instead of `AVALON_INSTANCE_ID` and `AVALON_CONTAINER_ID` (constants from ayon-core).
- [ ] If there are metadata stored to scene, we're referring to them as "ayon" metadata and not "openpype", "pype" or "avalon".
- [ ] If there are metadata related to folder, it is not referred to it as "asset" (OpenPype term) - Use correct folder path or folder name based on functionality.
- [ ] If there are metadata related to product, it is not referred to it as "subset" (OpenPype term)
- [ ] If there are metadata related to product type, it is not referred to it as "family", except for publishing (OpenPype term)

And codebase need to change any of that, we must make sure it can be backwards compatible (read older scenes).

Create separate issues if it should be split to multiple PRs, and close the issue if all conditions are met.

PR #115, opened 2025-05-09
Replace avalon and openpype related reference with ayon related reference

moonyuet:
## Changelog Description
This PR is to replace avalon and openpype related reference with ayon-related reference and implement the backward compatibility of these deprecated attributes.
Fix https://github.com/ynput/ayon-blender/issues/30

> [!IMPORTANT] 
> All the avalon-related loaded assets and instances would be deprecated after this PR has been merged. 

n/a

## Testing notes:
1. Launch Blender
2. Test Everything(Publish, load, manage etc)


client/ayon_blender/api/lib.py
```diff
@@ -252,19 +252,23 @@ def lsattrs(attrs: Dict) -> List:
...
     return list(matches)
 
 
 def read(node: bpy.types.bpy_struct_meta_idprop):
     """Return user-defined attributes from `node`"""
 
-    data = dict(node.get(pipeline.AVALON_PROPERTY, {}))
+    data = dict(node.get(pipeline.AYON_PROPERTY, {}))
+    if node.get(pipeline.AVALON_PROPERTY, {}):
+        data.update(dict(node.get(pipeline.AVALON_PROPERTY, {})))
```
BigRoy, line 271: This would **always** make any avalon property override any ayon property.
I'd say, if we want that - we should do the reverse and have AYON override the legacy one.
But better yet most likely is to only get the AVALON property data if there is no AYON property data.

client/ayon_blender/api/ops.py
```diff
@@ -250,6 +250,7 @@ def execute(self, context):
             if window is None:
                 window = host_tools.get_tool_by_name(self._tool_name)
                 BlenderApplication.store_window(self.bl_idname, window)
+            print("window", self.bl_idname)
```
BigRoy, line 253:
```suggestion
```

client/ayon_blender/api/pipeline.py
```diff
@@ -155,6 +159,8 @@ def update_context_data(self, data: dict, changes: dict):
             changes (dict): Only data that has been changed. Each value has
                 tuple with '(<old>, <new>)' value.
         """
+        # Store in both properties for backward compatibility during transition
```
BigRoy, line 162: Can we mark in the comment at what time we think this should be removed? Like, say in three months from now? Or six? or version X or Y?

client/ayon_blender/api/pipeline.py
```diff
@@ -460,25 +466,27 @@ def _discover_gui() -> Optional[Callable]:
...
+    """Add the container to the Ayon container."""
 
-    avalon_container = bpy.data.collections.get(AVALON_CONTAINERS)
-    if not avalon_container:
-        avalon_container = bpy.data.collections.new(name=AVALON_CONTAINERS)
+    ayon_container = bpy.data.collections.get(AYON_CONTAINERS) or (
+        bpy.data.collections.get("AVALON_CONTAINERS")
+    )
+    if not ayon_container:
+        ayon_container = bpy.data.collections.new(name=AYON_CONTAINERS)
```
BigRoy, line 476: Technically - this would make it non-backwards compatible in the way that if one wanted to downgrade, these containers wouldn't get picked up. Not sure how strict we need to be there? @philippe-ynput @iLLiCiTiT ?

client/ayon_blender/api/pipeline.py
```diff
@@ -488,12 +496,17 @@ def metadata_update(node: bpy.types.bpy_struct_meta_idprop, data: Dict):
...
 
-    if not node.get(AVALON_PROPERTY):
-        node[AVALON_PROPERTY] = dict()
+    if not node.get(AYON_PROPERTY) and not node.get(AVALON_PROPERTY):
+        node[AYON_PROPERTY] = dict()
+
+    ayon_property_name = (
+        AYON_PROPERTY if not node.get(AVALON_PROPERTY)
+        else AVALON_PROPERTY
+    )
```
BigRoy, line 505: This should be reversed?

If AYON Property is present, it should be used.. it should not use the OLDER key if BOTH were present, right?

client/ayon_blender/api/plugin.py
```diff
@@ -53,18 +54,21 @@ def get_unique_number(
     folder_name: str, product_name: str
 ) -> str:
     """Return a unique number based on the folder name."""
-    avalon_container = bpy.data.collections.get(AVALON_CONTAINERS)
-    if not avalon_container:
+    ayon_container = bpy.data.collections.get(AYON_CONTAINERS) or (
+        bpy.data.collections.get("AVALON_CONTAINERS")
```
BigRoy, line 58: Why do we suddenly hardcode this value here instead of still using the `AVALON_CONTAINERS` constant? Seems like we're still better of keeping that constant in use?

BigRoy (changes requested):
Commented some remarks - I haven't reviewed all files because I noticed some of my comments started applying to most of the files - so I'll let you first refactor before continuing a review.

LiborBatek:
My first finding during testing that former products (published before this change aka using the old ```AVALON``` string) are not listed in the ```Scene Inventory``` when loaded into active scene...

image

LiborBatek:
```
  File "C:\Work\REPO\ayon-core\client\ayon_core\tools\sceneinventory\models\containers.py", line 356, in _update_cache
    containers = list(host.get_containers())
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\pipeline.py", line 641, in ls
    yield parse_container(container)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\pipeline.py", line 621, in parse_container
    schema.validate(data)
  File "C:\Work\REPO\ayon-core\client\ayon_core\pipeline\schema\__init__.py", line 56, in validate
    jsonschema.validate(data,
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\jsonschema\validators.py", line 541, in validate
    cls(schema, *args, **kwargs).validate(instance)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\jsonschema\validators.py", line 130, in validate
    raise ERROR
jsonschema.exceptions.ValidationERROR: 'pyblish.avalon.instance' is not one of [ 'ayon.load.container' ]

FAILED validating 'enum' in schema[ 'properties']['id' ]:
```

LiborBatek:
Also when publishing asset aka creating ```publish instance``` it correctly adds custom property as ```ayon``` but not correctly setting up the collection with publish instances having it still as ```AVALON``` see pic below:

Screenshot 2025-05-12 180907

client/ayon_blender/api/lib.py
```diff
@@ -252,19 +252,24 @@ def lsattrs(attrs: Dict) -> List:
             continue
         for node in getattr(bpy.data, coll):
             for attr, value in attrs.items():
-                avalon_prop = node.get(pipeline.AVALON_PROPERTY)
-                if not avalon_prop:
+                ayon_prop = node.get(pipeline.AVALON_PROPERTY) or (
+                    node.get(pipeline.AYON_PROPERTY)
+                )
```
BigRoy, line 257: I think your reversed this the wrong way, no? I'm not sure how you interpreted my comment this way.

I'd say **AYON** property takes precedence over _any_ backwards compatibility.
As such, if there's any AYON property - that should be read first.

client/ayon_blender/api/lib.py
```diff
@@ -252,19 +252,24 @@ def lsattrs(attrs: Dict) -> List:
...
+                if not ayon_prop:
                     continue
```
BigRoy, line 259: Looking at this code - this logic is better moved to outside the `for attr, value ...` loop because it's not dependent on it at all.

client/ayon_blender/api/lib.py
```diff
@@ -243,28 +243,43 @@ def lsattrs(attrs: Dict) -> List:
...
+    ]
+
+    for coll in collections:
+        for node in coll:
+            # AYON property precedes over the AVALON one
+            ayon_prop = node.get(pipeline.AYON_PROPERTY) or node.get(pipeline.AVALON_PROPERTY)
+            if not ayon_prop:
+                continue
+
+            # Check all  key attributes which are included in the ayon property data
```
BigRoy, line 265:
```suggestion
            # Check all key attributes which are included in the ayon property data
```

client/ayon_blender/api/lib.py
```diff
@@ -243,28 +243,43 @@ def lsattrs(attrs: Dict) -> List:
...
 
 
 def read(node: bpy.types.bpy_struct_meta_idprop):
     """Return user-defined attributes from `node`"""
 
-    data = dict(node.get(pipeline.AVALON_PROPERTY, {}))
+    data = dict(
+        node.get(pipeline.AVALON_PROPERTY, {}) or
+        node.get(pipeline.AYON_PROPERTY, {})
+    )
```
BigRoy, line 282:
```suggestion
    data = dict(
        node.get(pipeline.AYON_PROPERTY, {}) or
        node.get(pipeline.AVALON_PROPERTY, {})
    )
```

client/ayon_blender/api/pipeline.py
```diff
@@ -142,6 +146,8 @@ def get_context_data(self) -> dict:
             dict: Context data stored using 'update_context_data'.
         """
         property = bpy.context.scene.get(AVALON_PROPERTY)
+        if not property:
+            property = bpy.context.scene.get(AYON_PROPERTY)
```
BigRoy, line 150:
```suggestion
        property = bpy.context.scene.get(AYON_PROPERTY)
        if not property:
            property = bpy.context.scene.get(AVALON_PROPERTY)
```

client/ayon_blender/api/pipeline.py
```diff
@@ -155,6 +161,9 @@ def update_context_data(self, data: dict, changes: dict):
             changes (dict): Only data that has been changed. Each value has
                 tuple with '(<old>, <new>)' value.
         """
+        # Store in both properties for backward compatibility during
+        # three months after this implementation
```
BigRoy, line 165:
```suggestion
        # Store in both properties for backward compatibility
        # TODO: Remove AVALON_PROPERTY persisting by September 2025
```

client/ayon_blender/api/pipeline.py
```diff
@@ -460,25 +469,27 @@ def _discover_gui() -> Optional[Callable]:
     return None
 
 
-def add_to_avalon_container(container: bpy.types.Collection):
-    """Add the container to the Avalon container."""
+def add_to_ayon_container(container: bpy.types.Collection):
+    """Add the container to the Ayon container."""
```
BigRoy, line 473:
```suggestion
    """Add the container to the AYON container."""
```

client/ayon_blender/api/pipeline.py
```diff
@@ -460,25 +469,27 @@ def _discover_gui() -> Optional[Callable]:
...
 
-    avalon_container = bpy.data.collections.get(AVALON_CONTAINERS)
-    if not avalon_container:
-        avalon_container = bpy.data.collections.new(name=AVALON_CONTAINERS)
+    ayon_container = bpy.data.collections.get(AVALON_CONTAINERS) or (
```
BigRoy, line 475:
```suggestion
    # Prioritize the legacy name here so that in existing scenes, if user would downgrade
    # to an older release the new data would also be in the legacy container.
    ayon_container = bpy.data.collections.get(AVALON_CONTAINERS) or (
```

client/ayon_blender/api/pipeline.py
```diff
@@ -487,13 +498,14 @@ def metadata_update(node: bpy.types.bpy_struct_meta_idprop, data: Dict):
 
     Existing metadata will be updated.
     """
-
-    if not node.get(AVALON_PROPERTY):
-        node[AVALON_PROPERTY] = dict()
+    ayon_property_name = (
+        AVALON_PROPERTY if not node.get(AYON_PROPERTY)
+        else AYON_PROPERTY
+    )
```
BigRoy, line 504:
```suggestion
    # TODO: By September 2025 we can remove the AVALON_PROPERTY
    #  fallback and only write into the new property
    ayon_property_name = (
        AYON_PROPERTY if node.get(AYON_PROPERTY)
        else AVALON_PROPERTY
    )
```

client/ayon_blender/api/pipeline.py
```diff
@@ -530,17 +542,33 @@ def containerise(name: str,
...
+    ayon_schema = (
+        "openpype:container-2.0"
+        if container.get(AVALON_PROPERTY)
+        else "ayon:container-3.0"
+    )
+    container_id = (
+        AVALON_CONTAINERS
+        if container.get(AVALON_PROPERTY)
+        else AYON_CONTAINERS
+    )
```
BigRoy, line 559: This logic makes no sense :)... We're creating a NEW collection. The container object will hence NEVER have any of these attributes.

I think we either need to add a `todo` if for now we want to keep the old ids/schemas OR just start using the new ones here directly.

client/ayon_blender/api/pipeline.py
```diff
@@ -569,9 +597,25 @@ def containerise_existing(
...
```
BigRoy, line 614:
```suggestion
    container_id = AYON_CONTAINERS
    ayon_schema = "ayon:container-3.0"
    
    # Backwards compatibility for legacy containers
    if container.get(AVALON_PROPERTY)
        container_id = AVALON_CONTAINERS
        ayon_schema = "openpype:container-2.0"
```

client/ayon_blender/api/pipeline.py
```diff
@@ -631,10 +675,11 @@ def ls() -> Iterator:
     node_tree = bpy.context.scene.node_tree
     if node_tree:
         for node in node_tree.nodes:
-            if not node.get(AVALON_PROPERTY):
+            prop_data = node.get(AVALON_PROPERTY) or node.get(AYON_PROPERTY)
```
BigRoy, line 678:
```suggestion
            prop_data = node.get(AYON_PROPERTY) or node.get(AVALON_PROPERTY)
```

client/ayon_blender/api/pipeline.py
```diff
@@ -646,10 +691,13 @@ def ls() -> Iterator:
             continue
 
         for shader_node in material_node_tree.nodes:
-            if not shader_node.get(AVALON_PROPERTY):
+            prop_data = shader_node.get(AVALON_PROPERTY) or (
+                shader_node.get(AYON_PROPERTY)
+            )
```
BigRoy, line 696:
```suggestion
            prop_data = shader_node.get(AYON_PROPERTY) or (
                shader_node.get(AVALON_PROPERTY)
            )
```

BigRoy (changes requested):
I think there was still some misunderstanding.

When reading the properties, if AYON property is present it should be used over any AVALON property because a newer value would be more important than the old.

When writing, if an AVALON property happens to exist we should solely write into that if we want to maintain backwards compatibility and support 'downgrading' the addon.

We can also say, we just don't care about downgrading the addon and just "force" everything to the new format. Looking at the amount of code involved in this PR _just to support both_ even though no tests are included it is **actually** compatible I'd personally like to argue to ditch backwards compatibility in the way that _downgrading_ a scene to an older `ayon-blender` release would not work - but we only support "updating" legacy scenes to work with this refactored state.

That could massively take down the scope that's needed here with all the "if this or that" branching.

Now suddenly all loaders and creators start trying to read/write both properties. Seems like we should abstract that away much better so that any compatibility is just a call to:

```
container = get_ayon_container()
```

Instead of everywhere doing:
```
        ayon_container = bpy.data.collections.get(AVALON_CONTAINERS) or (
            bpy.data.collections.get(AYON_CONTAINERS)
        )
```
etc.

@antirotor @philippe-ynput thoughts?

client/ayon_blender/plugins/create/create_animation.py
```diff
@@ -6,7 +6,7 @@
 class CreateAnimation(plugin.BlenderCreator):
     """Animation output for character rigs."""
 
-    identifier = "io.openpype.creators.blender.animation"
+    identifier = "io.ayon.creators.blender.animation"
```
BigRoy, line 9: Coming back to [this bigger comment](https://github.com/ynput/ayon-blender/pull/115#pullrequestreview-2839828224) does updating the identifiers now mean that any newly created instance would not be detected if one _were_ to downgrade? (Not saying that's critical, but if that would not be supported (and we think that's fine) then I think we can definitely start removing anything that's trying to be supposedly backwards compatible.

Might need an opinion from @iLLiCiTiT @philippe-ynput @antirotor however.

BigRoy: It does mean that indeed - however, Philippe stated that's totally ok. Downgrading to older versions introducing incompatibilities is fine if the versioning scheme at least shows that this is a bigger change, e.g. minor or major version bump.

BigRoy (commented):
I may be missing something - but where's the logic that converts legacy scenes to the AYON property format so that they are forward compatible?

A quick read over the code makes me think that now loading a legacy scene does not work at all?

The way I understood from @philippe-ynput was that the goal would become to:
- Whenever we start querying and persisting this data, to convert it to the new property. So that those 'entry points' would already convert things for us so that we'd need only a few places where this "upgrading" conversion would occur instead of needing to look for both AVALON and AYON property in both cases. I just can't seem to find where we now implementing that upgrade conversion for the avalon -> ayon property?

moonyuet:
> I may be missing something - but where's the logic that converts legacy scenes to the AYON property format so that they are forward compatible?
> 



> I may be missing something - but where's the logic that converts legacy scenes to the AYON property format so that they are forward compatible?
> 
> A quick read over the code makes me think that now loading a legacy scene does not work at all?
> 
> The way I understood from @philippe-ynput was that the goal would become to:
> 
> * Whenever we start querying and persisting this data, to convert it to the new property. So that those 'entry points' would already convert things for us so that we'd need only a few places where this "upgrading" conversion would occur instead of needing to look for both AVALON and AYON property in both cases. I just can't seem to find where we now implementing that upgrade conversion for the avalon -> ayon property?

Okay gotcha. I will do it and tag you for review. But review with the comment if it is wrong.

client/ayon_blender/api/pipeline.py
iLLiCiTiT: You're right, but I'm afraid that we can't really make it backwards compatible.

NOTE Release with this PR will raise version to `1.0.0`.

client/ayon_blender/api/lib.py
```diff
@@ -251,20 +251,26 @@ def lsattrs(attrs: Dict) -> List:
...
+            if not ayon_prop:
                 avalon_prop = node.get(pipeline.AVALON_PROPERTY)
                 if not avalon_prop:
                     continue
-                if (avalon_prop.get(attr)
-                        and (value is None or avalon_prop.get(attr) == value)):
+                else:
+                    node[pipeline.AYON_PROPERTY] = avalon_prop
+                    ayon_prop = avalon_prop
+                    del avalon_prop
```
iLLiCiTiT, line 262: This is just guess, but I think it should be.
```suggestion
                    del node[pipeline.AVALON_PROPERTY]
```

NOTE: Could be the constants imported to avoid `pipeline.{CONSTANT}`?

client/ayon_blender/api/pipeline.py
```diff
@@ -646,10 +654,21 @@ def ls() -> Iterator:
...
+                avalon_shader_node = shader_node.get(
+                    AVALON_PROPERTY)
+                if not avalon_shader_node:
+                    continue
+                else:
+                    ayon_shader_node[AYON_PROPERTY] = (
+                        avalon_shader_node
+                    )
+                    ayon_shader_node = avalon_shader_node
+                    del ayon_shader_node[AVALON_PROPERTY]
```
iLLiCiTiT, line 669: No need for else if there is `continue`.
```suggestion
                ayon_shader_node[AYON_PROPERTY] = (
                    avalon_shader_node
                )
                ayon_shader_node = avalon_shader_node
                del ayon_shader_node[AVALON_PROPERTY]
```

client/ayon_blender/api/plugin.py
```diff
@@ -295,7 +299,7 @@ def collect_instances(self):
 
         # Process only instances that were created by this creator
         for instance_node in cached_instances.get(self.identifier, []):
-            property = instance_node.get(AVALON_PROPERTY)
+            property = instance_node.get(AYON_PROPERTY)
```
iLLiCiTiT, line 302: This is guess, shouldn't this also have the fallback?
```suggestion
            property = instance_node.get(AYON_PROPERTY)
            if not propery:
                property = instance_node.get(AVALON_PROPERTY)
```

iLLiCiTiT:
Workfile saved with this PR makes the workfile backwards incompatible with older addon versions.
- We must raise major version with this PRa and add it to the release notes when done.

iLLiCiTiT:
We should add new converter to convert instances with `openpype` identifiers to those with `ayon` -> or do it automatically during `collect_instances`, since we already do automated changes at other places.

client/ayon_blender/api/ops.py
```diff
@@ -471,10 +471,10 @@ def draw(self, context):
         layout.separator()
         layout.operator(LaunchWorkFiles.bl_idname, text="Work Files...")
 
-def draw_avalon_menu(self, context):
-    """Draw the Avalon menu in the top bar."""
+def draw_ayon_menu(self, context):
+    """Draw the Ayon menu in the top bar."""
```
iLLiCiTiT, line 475:
```suggestion
    """Draw the AYON menu in the top bar."""
```

client/ayon_blender/api/pipeline.py
iLLiCiTiT: It is missing feature in the CI, looks like it would require changes in at about 3 CI actions. But the release can be done by manually passing new version in release trigger action.

client/ayon_blender/api/plugin.py
```diff
@@ -202,37 +203,40 @@ def cache_instance_data(shared_data):
...
+
+                if ayon_prop.get('id') not in {
                     AYON_INSTANCE_ID, AVALON_INSTANCE_ID
                 }:
                     continue
 
-                creator_id = avalon_prop.get('creator_identifier')
-                if creator_id:
+                creator_id = ayon_prop.get('creator_identifier')
+                if "openpype" not in creator_id:
```
iLLiCiTiT, line 228:
```suggestion
                creator_id = ayon_prop.get('creator_identifier')
                if "openpype" in creator_id:
                    creator_id = creator_id.replace("openpype", "ayon")

                if creator_id:
```

client/ayon_blender/api/plugin.py
```diff
@@ -202,37 +203,40 @@ def cache_instance_data(shared_data):
             cache = {}
             cache_legacy = {}
 
-            avalon_instances = bpy.data.collections.get(AVALON_INSTANCES)
-            avalon_instance_objs = (
-                avalon_instances.objects if avalon_instances else []
+            ayon_instances = bpy.data.collections.get(AYON_INSTANCES)
```
iLLiCiTiT, line 206: Shouldn't we relink avalon instances into ayon instances and remove avalon collection afterwards?

BigRoy: Yes, the idea should be that we "convert" the old data into the new data (early on) so that we can rely in most areas of the code on the new `AYON_INSTANCES` so we do not need to check for both everywhere.

iLLiCiTiT: Is this part already implemented?

client/ayon_blender/api/pipeline.py
```diff
@@ -460,25 +466,42 @@ def _discover_gui() -> Optional[Callable]:
...
+
 
-    avalon_container = bpy.data.collections.get(AVALON_CONTAINERS)
-    if not avalon_container:
-        avalon_container = bpy.data.collections.new(name=AVALON_CONTAINERS)
+def convert_avalon_instances():
+    avalon_instances = bpy.data.collections.get(AVALON_INSTANCES)
+    if not avalon_instances:
+        return
+    avalon_instances.name = AYON_INSTANCES
```
BigRoy, line 483: What if `AYON_INSTANCES` already exists as well?

LiborBatek:
When tested Im getting ```No asset group found``` when trying to load any Ayon product via ```Loader```

```

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\ops.py", line 129, in execute
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\plugin.py", line 499, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\plugins\load\load_blend.py", line 152, in process_asset
    container, members = self._process_data(libpath, group_name)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\plugins\load\load_blend.py", line 95, in _process_data
    assert container, "No asset group found"
           ^^^^^^^^^

AssertionError: No asset group found

```

will post more findings in separate post as I will progress with testing...

LiborBatek:
Im getting errors while trying to load any products:

```Traceback (most recent call last):

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\ops.py", line 129, in execute
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\plugin.py", line 499, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\plugins\load\load_cache.py", line 180, in process_asset
    convert_avalon_containers()

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.10+dev\ayon_blender\api\pipeline.py", line 527, in convert_avalon_containers
    avalon_containers.name = AYON_CONTAINERS
    ^^^^^^^^^^^^^^^^^^^^^^

AttributeError: 'NoneType' object has no attribute 'name'
```

client/ayon_blender/api/plugin.py
iLLiCiTiT: Shouldn't `convert_avalon_instances` be called before `ayon_instances` are received?

LiborBatek (changes requested):
Gives me error while trying to load (Append) ```Rig``` product

```

             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\plugin.py", line 498, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_blend.py", line 160, in process_asset
    container, members = self._process_data(libpath, group_name)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

LiborBatek (approved):
Now all looks good to me!

As far as I can tell...been testing publishing different products and load those and opening workfiles etc.


all seems  properly! <p style="color:MediumSeaGreen;">Ayonized™</p>

client/ayon_blender/api/pipeline.py
```diff
@@ -438,25 +444,67 @@ def _discover_gui() -> Optional[Callable]:
...
 
-def add_to_avalon_container(container: bpy.types.Collection):
-    """Add the container to the Avalon container."""
+def get_ayon_property_by_avalon_property(node):
+    property = node.get(AYON_PROPERTY)
+    if not property:
+        property = node.get(AVALON_PROPERTY)
+        if property:
+            node[AYON_PROPERTY] = property
+            del node[AVALON_PROPERTY]
```
BigRoy, line 453: I wonder if this should log a debug message.

BigRoy (approved):
**Had some remarks**, see new comments. Other than that it looks good.

The only issue I can think of is loading _legacy publishes_ using linking if that loaded data has nodes with the legacy avalon property that the code may error due to NOT being able to manipulate the property on the node due to it being linked/non-editable. However, that's an edge case we probably want to solve if we face it...

Just approving so that if those changes are made it can be merged perhaps.

philippe-ynput:
@BigRoy @moonyuet yes, we need a major bump.

Issue #71, opened 2024-11-08
Render review and thumbnail fail due to missing OCIO config


jrsndl:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
With path to OCIO config set in core settings: ```ayon+settings://core/imageio/ocio_config_profiles``` for Blender host,
Using Cycles renderer with compositing enabled fails to make review and thumbnail on farm.

The OIIOtool expects valid path to OCIO.

![Image](https://github.com/user-attachments/assets/a25fa861-daa8-4e2f-9339-130a311c55f3)

### Steps To Reproduce:

1. Configure OIIOtool to convert the exr to png sequence
2. Publish exr render from Blender 4.2 to Deadline
3. No review or thumbnail is created

### Version
Server 1.42, ayon_core 1.06, ayon_blender 0.2.5

### What platform were you running when you found the bug?
Windows

You can also provide any additional information relevant to the issue.

This change can be a quick fix:

- It uses OCIO path from envinronment (might be an issue for multi-platform farms)
- The hardcoded Display and View values are swapped and use more typical OCIO 1.0 names

![Image](https://github.com/user-attachments/assets/4feb2237-fbbe-48e7-8795-69b8d80ed162)

https://github.com/ynput/ayon-blender/blob/b9ea07ec0098a4ca00727e87a6b374f526a13e8e/client/ayon_blender/plugins/publish/collect_render.py#L110C13-L112C52

### Relevant log output:

```shell
INFO:pyblish.ExtractThumbnail:oiiotool ERROR: -ociodisplay : DisplayViewTransform error. Cannot find source color space named 'Linear'.
2024-11-08 10:59:38:  0: STDOUT: Full command line was:
[...]

PR #119, opened 2025-06-14
Collect OCIO path from environment

jrsndl:
## Changelog Description
Extract OIIO Transcode is not able to produce review from Blender, 'cause Blender doesn't collect the OCIO config path.

See [here](https://github.com/ynput/ayon-core/blob/f8ff18f3a76c026db280a12acb635e248e402118/client/ayon_core/plugins/publish/extract_color_transcode.py#L99)

This PR collects path from environment variable if present.

Can close https://github.com/ynput/ayon-blender/issues/71

## Testing notes:
1. Configure OIIOtool to convert the exr to png sequence
2. Enable Blender color management
3. Publish exr render from Blender 4.2 to Deadline
4. Check if review is created

### DIFF
--- a/client/ayon_blender/plugins/publish/collect_render.py
+++ b/client/ayon_blender/plugins/publish/collect_render.py
@@ -86,7 +86,7 @@ class CollectBlenderRender(plugin.BlenderInstancePlugin):
                 # OCIO not currently implemented in Blender, but the following
                 # settings are required by the schema, so it is hardcoded.
                 # TODO: Implement OCIO in Blender
-                "colorspaceConfig": "",
+                "colorspaceConfig": os.environ.get("OCIO", ""),
                 "colorspaceDisplay": "sRGB",
                 "colorspaceView": "ACES 1.0 SDR-video",
                 "renderProducts": colorspace.ARenderProduct(


BigRoy:
@jrsndl could you try this instead? https://github.com/jrsndl/ayon-blender/pull/1
That would at least get the OCIO config from the scene correctly - otherwise the display and view would be hardcoded to potentially the wrong value.

However, technically this value is also still hardcoded: https://github.com/ynput/ayon-blender/blob/ba1ba1c8c9fd33522df6518488aea9d133f02a91/client/ayon_blender/api/colorspace.py#L47-L48 and likely more important to get right.

Maybe time is better spend on getting #67 through asap - which would fix this, and provide a lot more use cases.

jrsndl:
> @jrsndl could you try this instead? [jrsndl#1](https://github.com/jrsndl/ayon-blender/pull/1)
> That would at least get the OCIO config from the scene correctly - otherwise the display and view would be hardcoded to potentially the wrong value.

Hi, I've tested the Backport get_colorspace_data PR, it is working and much better than just fixing the OCIO path.  I will close this PR...
Thank you @BigRoy

Issue #121, opened 2025-06-20
Loading model (FBX) into active scene triggers error message


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
When trying to load ```Model``` product ```FBX``` it triggers error however the product get loaded.

![Image](https://github.com/user-attachments/assets/cfd7e68f-50be-4608-87e4-345b69f813e4)

```
Traceback (most recent call last):

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.11\ayon_blender\api\ops.py", line 129, in execute
    result = callback(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.11\ayon_blender\api\plugin.py", line 495, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.11\ayon_blender\plugins\load\load_fbx.py", line 152, in process_asset
    objects = self._process(libpath, asset_group, group_name, None)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_0.2.11\ayon_blender\plugins\load\load_fbx.py", line 90, in _process
    parent.objects.link(obj)

RuntimeError: Error: Object 'model' already in collection 'Scene Collection'

```

## Expected Behavior
There shouldnt be any error msg while using FBX repre for loading the product.

### Steps To Reproduce:

1. Pick ```model``` with ```FBX``` repre
2. Load (FBX) into active workfile
3. Error should occur.

### Version
```ayon-blender``` addon ```0.2.11```

### What platform were you running when you found the bug?
Windows

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #122, opened 2025-06-20
Load fbx: Checking objects already in scene collection during loading

moonyuet:
## Changelog Description
This PR is to ensure scene collection would not link duplicate active objects in the scene collection thus gives error.
Resolves https://github.com/ynput/ayon-blender/issues/121

n/a

## Testing notes:
1. Load fbx
2. Should be working

### DIFF
--- a/client/ayon_blender/plugins/load/load_fbx.py
+++ b/client/ayon_blender/plugins/load/load_fbx.py
@@ -59,6 +59,8 @@ class FbxModelLoader(plugin.BlenderLoader):
         if blender_version >= (4, 5, 0):
             bpy.ops.wm.fbx_import(filepath=libpath)
         else:
+            # TODO: make sure it works with the color management
+            # in 4.4 or elder version
             bpy.ops.import_scene.fbx(filepath=libpath)
 
         parent = bpy.context.scene.collection


moonyuet:
@LiborBatek I got some update on the fbx loading issue in regard to the color management. The issue is coming from our ocio config does not include `Non-Color` as being part of colorspace attributes.
This can be temporarily fixed by adding ocio config path which is **specifically designed** for Blender.
You can find it through `C:\Program Files\Blender Foundation\Blender {YOUR_BLENDER_VERSION}\{YOUR_BLENDER_VERSION}\datafiles\colormanagement`
image

@jakubjezek001 @rdelillo  @BigRoy should we add this blender-specific attribute into the OCIO or?

BigRoy:
> should we add this blender-specific attribute into the OCIO or?

No. Because we cannot rely on others having it in their custom OCIO files either.
We should solve what's the root of the issue instead.

Why does the FBX contain "Non-Color" information? Where does that originate from?
Or is it not that the FBX contains "Non-Color" but it's just really really really what FBX Importer of Blender wants to set by default?

pleprince:
@moonyuet @LiborBatek You should NEVER try to find a colorspace name. You need to find a [role](https://opencolorio.readthedocs.io/en/latest/guides/authoring/overview.html#config-roles) in the OCIO config. The `role` maps a colorspace usage to an actual colorspace in the current config.
In this case, it seems that a blender\fbx flag named `Non-Color` should point to the `data` role, which itself will point the correct colorspace.

LiborBatek (approved):
I have sucessfully tested both in ```Blender 4.4.3``` and ```Blender 4.5.0``` and all working as expected. I was also able to version manage the loaded ```FBX``` repre as with other product types.

LGTM
Screenshot 2025-06-30 133529

Issue #116, opened 2025-05-29
Unsupported type Image Texture


m-u-r-p-h-y:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
error in publishing introduced in 0.2.11

### Version
Blender 0.2.11

You can also provide any additional information relevant to the issue.

![Image](https://github.com/user-attachments/assets/5d969a25-d54c-44c6-9d4f-3904f96136e7)

### Relevant log output:

```shell

```

alainxi:
Hi,

I can reproduce the bug every time :

Make sure you publish an image from some DCC (in my case I published textures from Substance).

In Blender, create a Cube, and select the Cube.

Ayon-->Create-->  PublishType=Model, UseSelection.

Ayon-->Load-->SelectAnImage-->Load to shader editor

In the Shading tab of Blender, connect the "Image Texture" node (that was created by "Load to shader editor") to the Base Color of the Principled BSDF.

Ayon--Publish


Here are the publish errors :

```
File "C:\Users\Alain\AppData\Local\Ynput\AYON\addons\blender_0.2.12\ayon_blender\api\lib.py", line 625, in strip_namespace
    raise TypeError(f"Unsupported type: {node} ({type(node)})")
TypeError: Unsupported type: <bpy_struct, ShaderNodeTexImage("Image Texture") at 0x0000021AFD679608> (<class 'bpy.types.ShaderNodeTexImage'>)
```


Here is the Blender scene, that I saved after the publish failed :

[WOLF_mod_v002.zip](https://github.com/user-attachments/files/21161010/WOLF_mod_v002.zip)

PR #124, opened 2025-06-23
Supports stripping namespace for shader nodes for texture image

moonyuet:
## Changelog Description
This PR is to add support for striping namespace for shader nodes with texture images when extracting blend scene
Fixes https://github.com/ynput/ayon-blender/issues/116

n/a

## Testing notes:
1. Create Rig/BlendScene with shader nodes(texture-image based) 
2. Publish

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -621,6 +621,8 @@ def strip_namespace(containers):
             children = node.children_recursive
         elif isinstance(node, bpy.types.Object):
             children = node.children
+        elif isinstance(node, bpy.types.ShaderNodeTexImage):
+            children = [node]
         else:
             raise TypeError(f"Unsupported type: {node} ({type(node)})")
 
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -621,7 +621,7 @@ def strip_namespace(containers):
             children = node.children_recursive
         elif isinstance(node, bpy.types.Object):
             children = node.children
-        elif isinstance(node, bpy.types.ShaderNodeTexImage):
+        elif isinstance(node, bpy.types.Node):
             children = [node]
         else:
             raise TypeError(f"Unsupported type: {node} ({type(node)})")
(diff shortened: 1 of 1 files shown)


client/ayon_blender/api/lib.py
```diff
@@ -621,6 +621,8 @@ def strip_namespace(containers):
             children = node.children_recursive
         elif isinstance(node, bpy.types.Object):
             children = node.children
+        elif isinstance(node, bpy.types.ShaderNodeTexImage):
```
BigRoy, line 624: Is there's a reason to not just support the more generic base class `ShaderNode` or even `Node`?

BigRoy (approved):
Code looks good to me. @LiborBatek can you test?

LiborBatek:
Couldnt test it, provide more exact testing condition/ test steps...

alainxi:
Hi,

I provided the steps to reproduce the bug, and the Blender scene, here :

https://github.com/ynput/ayon-blender/issues/116#issuecomment-3056844209

alainxi:
- First try :   click on Reset in the Publish gui of the Blender scene that was still opened after failure.    The same error occurs.
- Second try : delete the instance, create again the same instance.    Now the publish works all right !

As it's an Extract, I thought it would have worked in both tries.

moonyuet:
@alainxi can you try again with the latest commit [67f3309](https://github.com/ynput/ayon-blender/pull/124/commits/67f3309909066fcf5b20054bbd4e2586a807b3a8) ? Should be working with your two tries.

moonyuet:
> Actually this should have been absolutely redundant - and should not have fixed it...
> 
> The [ShaderNodeTexImage](https://docs.blender.org/api/current/bpy.types.ShaderNodeTexImage.html) inherits from `Node` which we check `isinstance` for JUST ABOVE? This seems very fishy.

Let's push another change for this and merge it again, when it makes no differences.

BigRoy:
Ah, nevermind - it's the `Node` you added as well - in that case the [ShaderNodeTexImage](https://docs.blender.org/api/current/bpy.types.ShaderNodeTexImage.html) can just be removed and only the `Node` check can be kept. In that case, the `Node` isinstance check would've fixed it.

moonyuet:
> I confirm that it's the `Node` line that fixed it for me.
> 
> The redondant `ShaderNodeTexImage` did not change anything in my tests (the results were the same, with or without it).

The PR #143 which removes the redundant code is already merged. Thank you for the testing!

Issue #127, opened 2025-07-02
Select hierarchies of the instances during publishing


moonyuet:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

The current publishing workflow needs users to manually select the hierarchy before extraction. 
If the users do not do that, it would publish some empty/missing assets.

### Suggested implementation?

We can write some recursive maintained_selection function which select the object and publish during context.
Some pesudo code can be taken as reference:
```python
import bpy

# Function to recursively select all children of an object
def select_hierarchy(obj):
    # Select the parent object
    obj.select_set(True)
    
    # Recursively select all children
    for child in obj.children:
        select_hierarchy(child)

```

moonyuet:
> ### Is there an existing issue for this?
> * [x]  I have searched the existing issues and added correct labels.
> 
> ### Please describe the feature you have in mind and explain what the current shortcomings are?
> The current publishing workflow needs users to manually select the hierarchy before extraction. If the users do not do that, it would publish some empty/missing scene.
> 
> ### Suggested implementation?
> We can write some recursive maintained_selection function which select the object and publish during context. Some pesudo code of selecting hierarchies can be taken as reference:
> 
> import bpy
> 
> # Function to recursively select all children of an object
> def select_hierarchy(obj):
>     # Select the parent object
>     obj.select_set(True)
>     
>     # Recursively select all children
>     for child in obj.children:
>         select_hierarchy(child)
> ### Describe alternatives you've considered:
> _No response_

BigRoy:
> The current publishing workflow needs users to manually select the hierarchy before extraction.
> If the users do not do that, it would publish some empty/missing assets.

Wait - what? For which extractor is this the case?
Did you mean that it's needed for **creation**?


I suppose this is about the fact that on certain exports, e.g. blend scene if the children are NOT in the publish collection, but only the roots then the children are not included in the export. 

Although unintuitive to Maya users perhaps, this may be expected for Blender users?
And if not, there may still be cases where you want to export a parent without the children, so an automatic "always include all children" during extraction doesn't seem right either.

moonyuet:
I didn't mean for creation.(Sorry for not providing too much info) We need to make sure select the hierarchies we want to publish during publishing and restore what we currently select(or deselect if we dont select) afterwards.  This is needed to be implemented in the extractor(s) with context manager, like we did similarly in Maya.

[...]

PR #128, opened 2025-07-03
 Collect the children recursive of the instances when instance node is Collection

moonyuet:
## Changelog Description
This PR is to use `children_recursive` to collect all the instances members for publishing, this can ensure no missing data or empty scene being published.
Resolve https://github.com/ynput/ayon-blender/issues/127

All the creators need to be tested
Loaders to be tested:
Load Cache
Load Camera Abc
Load Camera Fbx
Load fbx
Load Image Shader

## Testing notes:
1. Create any instance
2. Publish

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -313,7 +313,7 @@ def get_selected_collections():
 
 def get_selection(
         include_collections: bool = False,
-        include_object_children_recursive: bool = True
+        include_object_children_recursive: bool = False
     )-> List[Union[bpy.types.Object, bpy.types.Collection]]:
     """
     Returns a list of selected objects in the current Blender scene.
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -311,24 +311,42 @@ def get_selected_collections():
     return [id for id in ids if isinstance(id, bpy.types.Collection)]
 
 
-def get_selection(include_collections: bool = False) -> List[bpy.types.Object]:
+def get_selection(
+        include_collections: bool = False,
+        selected_hierarchies: bool = False
+    )-> List[Union[bpy.types.Object, bpy.types.Collection]]:
     """
     Returns a list of selected objects in the current Blender scene.
 
     Args:
         include_collections (bool, optional): Whether to include selected
         collections in the result. Defaults to False.
+        selected_hierarchies (bool): Whether to include all children
+        of selected objects.
 
     Returns:
-        List[bpy.types.Object]: A list of selected objects.
+        List[Union[bpy.types.Object,
+        bpy.types.Collection]]: Selected objects and optionally collections.
     """
-    selection = [obj for obj in bpy.context.scene.objects if obj.select_get()]
+    selection = set(obj for obj in bpy.context.scene.objects if obj.select_get())
 
     if include_collections:
-        selection.extend(get_selected_collections())
+        selection.update(get_selected_collections())
 
-    return selection
+    if selected_hierarchies:
+        def collect_hierarchy(obj, collected):
+            for child in obj.children:
+                if child not in collected:
+                    collected.add(child)
+                    collect_hierarchy(child, collected)
 
+        hierarchy_set = set()
+        for obj in selection:
+            collect_hierarchy(obj, hierarchy_set)
+
+        selection.update(hierarchy_set)
+
+    return list(selection)
 
 @contextlib.contextmanager
 def maintained_selection():
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -311,10 +311,7 @@ def get_selected_collections():
     return [id for id in ids if isinstance(id, bpy.types.Collection)]
 
 
-def get_selection(
-        include_collections: bool = False,
-        selected_hierarchies: bool = False
-    )-> List[Union[bpy.types.Object, bpy.types.Collection]]:
+def get_selection(include_collections: bool = False)-> List[Union[bpy.types.Object, bpy.types.Collection]]:
     """
     Returns a list of selected objects in the current Blender scene.
 
@@ -332,21 +329,21 @@ def get_selection(
 
     if include_collections:
         selection.update(get_selected_collections())
-
-    if selected_hierarchies:
-        def collect_hierarchy(obj, collected):
-            for child in obj.children:
-                if child not in collected:
-                    collected.add(child)
-                    collect_hierarchy(child, collected)
-
-        hierarchy_set = set()
-        for obj in selection:
-            collect_hierarchy(obj, hierarchy_set)
-
-        selection.update(hierarchy_set)
+    selection.update(get_object_children_recursive(selection))
 
     return list(selection)
+
+
+def get_object_children_recursive(objects: set[bpy.types.Object]) -> set[bpy.types.Object]:
+    """Return all object children of any objects in the inputs.
+
+    Any input that is not a `bpy.types.Object` is skipped."""
+    children = set()
+    for obj in objects:
+        if isinstance(obj, bpy.types.Object):
+            children.update(obj.children_recursive)
+    return children
+
 
 @contextlib.contextmanager
 def maintained_selection():
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -311,39 +311,6 @@ def get_selected_collections():
     return [id for id in ids if isinstance(id, bpy.types.Collection)]
 
 
-def get_top_objects_selections()-> List[bpy.types.Object]:
-    """Get the top hierarchies of the selections
-
-    Args:
-        selections (List[Union[bpy.types.Object]]): selected objects
-
-    Returns:
-        List[bpy.types.Object]: top objects
-    """
-    selections = get_selection()
-    top_parents = {
-        obj: get_top_parent(obj) for obj in selections
-    }
-    top_object = sorted(
-        selections, key=lambda o: (o == top_parents[o], o.name), reverse=True
-    )
-    return top_object
-
-
-def get_top_parent(obj: bpy.types.Object) -> bpy.types.Object:
-    """Get top parent with iteration
-
-    Args:
-        obj (bpy.types.Object): Object
-
-    Returns:
-        bpy.types.Object: Top object
-    """
-    while obj.parent:
-        obj = obj.parent
-    return obj
-
-
 def get_selection(
         include_collections: bool = False,
         include_object_children_recursive: bool = True
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -311,9 +311,42 @@ def get_selected_collections():
     return [id for id in ids if isinstance(id, bpy.types.Collection)]
 
 
+def get_top_objects_selections()-> List[bpy.types.Object]:
+    """Get the top hierarchies of the selections
+
+    Args:
+        selections (List[Union[bpy.types.Object]]): selected objects
+
+    Returns:
+        List[bpy.types.Object]: top objects
+    """
+    selections = get_selection()
+    top_parents = {
+        obj: get_top_parent(obj) for obj in selections
+    }
+    top_object = sorted(
+        selections, key=lambda o: (o == top_parents[o], o.name), reverse=True
+    )
+    return top_object
+
+
+def get_top_parent(obj: bpy.types.Object) -> bpy.types.Object:
+    """Get top parent with iteration
+
+    Args:
+        obj (bpy.types.Object): Object
+
+    Returns:
+        bpy.types.Object: Top object
+    """
+    while obj.parent:
+        obj = obj.parent
+    return obj
+
+
 def get_selection(
         include_collections: bool = False,
-        include_object_children_recursive: bool = False
+        include_object_children_recursive: bool = True
     )-> List[Union[bpy.types.Object, bpy.types.Collection]]:
     """
     Returns a list of selected objects in the current Blender scene.
@@ -328,7 +361,9 @@ def get_selection(
         List[Union[bpy.types.Object,
         bpy.types.Collection]]: Selected objects and optionally collections.
     """
-    selection = set(obj for obj in bpy.context.scene.objects if obj.select_get())
+    selection = {
+        obj for obj in bpy.context.scene.objects if obj.select_get()
+    }
 
     if include_collections:
         selection.update(get_selected_collections())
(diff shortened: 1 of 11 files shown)


LiborBatek (changes requested):
I have tested with ```blender 4.5``` and ```blendScene``` product type for publishing.

I have setup small asset consisting of mixture of standalone objects, parented objects to locator and Collections containing some objects...

So far my findings are...works well for most things but is not working for nested hierarchies below ```locator```

Here is my scene before publish (note the hierarchy in the outliner)

Screenshot 2025-07-04 100046

And once loaded as published ```blendScene``` product into scene (Append action)
it is obviously missing objects sitting below locator, the rest of entities seems present. Sorry for scene been 90 deg rotated compared to the first (didnt spot it soon enough)
Screenshot 2025-07-04 095937

moonyuet:
> As stated in [internal chat](https://discord.com/channels/517362899170230292/940164626648285234/1390640625560322060) I think this is the wrong approach.
> 
> > If the user puts in a root group initially, creates instance, and later adds a new child object... the user still expects that in the export.
> > We should NOT manipulate that we add ALL children to the collection on create.
> > We should just collect from the collection the objects (+ all their children) and then extract that.
> 
> The logic should be here:
> 
> https://github.com/ynput/ayon-blender/blob/6cbebd37a3662a33a7b431837f2ed7927c3b162f/client/ayon_blender/plugins/publish/collect_instance.py#L22-L37



> As stated in [internal chat](https://discord.com/channels/517362899170230292/940164626648285234/1390640625560322060) I think this is the wrong approach.
> 
> > If the user puts in a root group initially, creates instance, and later adds a new child object... the user still expects that in the export.
> > We should NOT manipulate that we add ALL children to the collection on create.
> > We should just collect from the collection the objects (+ all their children) and then extract that.
> 
> The logic should be here:
> 
> https://github.com/ynput/ayon-blender/blob/6cbebd37a3662a33a7b431837f2ed7927c3b162f/client/ayon_blender/plugins/publish/collect_instance.py#L22-L37

Get to the develop branch, and create the instance again like this video below

https://github.com/user-attachments/assets/c9dc84e0-5974-497b-8d6b-36f47ed27d9e

moonyuet:
This is with this branch
https://github.com/user-attachments/assets/54deb6e5-ba54-4c04-a714-709e2958e11b

BigRoy:
Here's my proposal @moonyuet https://github.com/ynput/ayon-blender/pull/134

LiborBatek (commented):
I do like the new ```Select Hierarchies``` btn and it solves the issue well.

Screenshot 2025-07-04 163910

once enabled it pretty much outputs everything in the hierarchy which is desirable and correct!

here is my product loaded into empty scene containing everything. no missing parts.



Screenshot 2025-07-04 163823

Now as @BigRoy  stated during our vid call...what happens when the user adds something new into his Collection as this wont be reflected automatically to publish instance collection holding objects for publishing...

there should be probably some kind of logic check during publishing?? as a kind of validator / Repair action for adding such entities into that publish preferably?

BigRoy:
> there should be probably some kind of logic check during publishing?? as a kind of validator / Repair action for adding such entities into that publish preferably?

I really feel like if we **always** expect the children to be there in the output data - that we shouldn't need to mess with the "state of the instance" in the workfile, but just enforce that data to be there in the ouput file. It's just _more stuff to manage?_

LiborBatek:
> 
> I really feel like if we **always** expect the children to be there in the output data - that we shouldn't need to mess with the "state of the instance" in the workfile, but just enforce that data to be there in the ouput file. It's just _more stuff to manage?_

Yeah I really dont mind how this will happen (to ensure all is included)- the simpler the better! ...but it should def. happen automatically without user interaction each time he updates the asset with new entities.

moonyuet:
> there should be probably some kind of logic check during publishing checking those greyed out entities not falling into publish Instance collection - kind of validator / Repair action for adding such entities into that publish preferably?

I agree. We need some validator to check if the users include geometry data along with its transform data.
I am also thinking some scenario: User can accidentally select the objects (e.g. Asset A) + Collections which contains Asset A and publish. And the selection has been once published and been loaded back to the new scene. There would be collections with Asset A and the duplicate of the asset A. We need to make sure no duplicate into the publish instance.

BigRoy:
> We need some validator to check if the users include geometry data along with its transform data.

I'm so confused by this comment every time you mention there are issues with transform data and geometry data. I don't see that happening here at all. It's just that certain parts of the hierarchy are not included in the export (or for some of them the actual issue is that they are not included on load, but well.. visually same issue)

> I am also thinking some scenario: User can accidentally select the objects (e.g. Asset A) + Collections which contains Asset A and publish. And the selection has been once published and been loaded back to the new scene. There would be collections with Asset A and the duplicate of the asset A. We need to make sure no duplicate into the publish instance in the validator

I don't see how we can sensibly do this - there may very well be a use case where a user _intentionally did it_. If the user ended up putting the wrong thing in the Collection that's just plain human error or intentional. If 90% of the time in Blender managing collections is really shit for the artist, then maybe Blender should improve that 😈 but so far it seems fine on my end? (ignoring the hierarchy bit here)

client/ayon_blender/api/lib.py
```diff
@@ -311,24 +311,42 @@ def get_selected_collections():
     return [id for id in ids if isinstance(id, bpy.types.Collection)]
 
 
-def get_selection(include_collections: bool = False) -> List[bpy.types.Object]:
+def get_selection(
+        include_collections: bool = False,
+        selected_hierarchies: bool = False
```
BigRoy, line 316: I think adding this here is an odd choice.

It's a function that gets the selection - the hierarchy below it.. is not your selection? :)
I think we're better of just having a separate `get_hierarchy` function (also see comment below)

client/ayon_blender/api/lib.py
```diff
@@ -311,24 +311,42 @@ def get_selected_collections():
...
-        selection.extend(get_selected_collections())
+        selection.update(get_selected_collections())
 
-    return selection
+    if selected_hierarchies:
+        def collect_hierarchy(obj, collected):
+            for child in obj.children:
+                if child not in collected:
+                    collected.add(child)
+                    collect_hierarchy(child, collected)
```
BigRoy, line 341: No need for recursion.

Just use `Object.children_recursive`.

However, do avoid `children` or `children_recursive` on `Collection` objects unless you want to also link all INNER collections - which I don't think is the case here you only want the object hierarchies. So, skip it on Collection types.

I think we're better of implementing a separate `get_object_children_recursive` function if we really need this abstracted to `lib.py`

Or if we really want this inside `get_selection`, then perhaps do implement the separate function and use it in here. But it would really just be a few lines of code:

```python
def get_object_children_recursive(objects: list[bpy.types.Object]) -> set[bpy.types.Object]:
    """Return all object children of any objects in the inputs.
    
    Any input that is not a `bpy.types.Object` is skipped."""
    children = set()
    for obj in objects:
        if isinstance(obj, bpy.types.Object):
            children.update(obj.children_recursive)
    return children
```

client/ayon_blender/plugins/create/create_blendScene.py
```diff
@@ -40,3 +45,12 @@ def remove_instances(self, instances):
...
 
             self._remove_instance_from_context(instance)
+
+    def get_pre_create_attr_defs(self):
+        defs = super().get_pre_create_attr_defs()
+        defs.extend([
+            BoolDef("selected_hierarchies",
+                    label="Select Hierarchies",
+                    default=False)
+        ])
```
BigRoy, line 55: Select hierarchies is confusingly named - it doesn't select anything.

- Include hierarchy
- Include all children   (this may be more accurate, because we're not actually including parent hierarchy?)
- Link all recursive children   (this may be more accurate in blender terminology)
- Link all recursive children of objects   (this may be most accurate, because we might not want to do it on Collections?)

Anyway, this 'fixes' what's linked in the Collection solely for `CreateBlendScene` but the issue is also present on `CreatePointcache`.

BigRoy (changes requested):
Needs fixing for CreatePointcache as well + code can be improved cosmetically.
Also, let's have this toggle enabled by default in the UI.

client/ayon_blender/api/lib.py
moonyuet: It's by no doubt that your implementation is more leveraging the blender API and it is much faster. 

And we can choose to implement the manual recursive function again if we find the Blender API is not stable(it has deprecated setting for some selections) to have more control over traversal.

client/ayon_blender/api/lib.py
```diff
@@ -311,23 +311,38 @@ def get_selected_collections():
...
+        List[Union[bpy.types.Object,
+        bpy.types.Collection]]: Selected objects and optionally collections.
     """
-    selection = [obj for obj in bpy.context.scene.objects if obj.select_get()]
+    selection = set(obj for obj in bpy.context.scene.objects if obj.select_get())
 
     if include_collections:
-        selection.extend(get_selected_collections())
+        selection.update(get_selected_collections())
+    selection.update(get_object_children_recursive(selection))
```
BigRoy, line 332: Please do not put this inside `get_selection`. This should be a separate function call instead for those who need the hierarchy.

Or if you really prefer it inside `get_selection` put it behind an optional argument `include_object_children_recursive=False`. But a separate function call outside of `get_selection` makes more sense to me.

LiborBatek:
When testing loaders, it works ok for ```animation``` and ```pointcache```

```Camera``` works fine for ```Alembic``` but fails when ```FBX``` used for loading with following error:

```Traceback (most recent call last):

             ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\api\plugin.py", line 495, in _load
    nodes = self.process_asset(
            ^^^^^^^^^^^^^^^^^^^

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_camera_fbx.py", line 108, in process_asset
    self._process(libpath, asset_group, group_name)

  File "C:\Work\REPO\ayon-blender\client\ayon_blender\plugins\load\load_camera_fbx.py", line 58, in _process
    parent.objects.link(obj)

RuntimeError: Error: Object 'Camera' already in collection 'Scene Collection'
```

client/ayon_blender/api/lib.py
```diff
@@ -311,23 +311,42 @@ def get_selected_collections():
...
+        include_object_children_recursive (bool, optional): Whether to include all
+        hierarchies of selected objects.
 
     Returns:
-        List[bpy.types.Object]: A list of selected objects.
+        List[Union[bpy.types.Object,
+        bpy.types.Collection]]: Selected objects and optionally collections.
     """
-    selection = [obj for obj in bpy.context.scene.objects if obj.select_get()]
+    selection = set(obj for obj in bpy.context.scene.objects if obj.select_get())
```
BigRoy, line 331:
```suggestion
    selection = {obj for obj in bpy.context.scene.objects if obj.select_get()}
```

BigRoy (approved):

client/ayon_blender/api/lib.py
```diff
@@ -311,23 +311,77 @@ def get_selected_collections():
...
+
+    Args:
+        obj (bpy.types.Object): Object
+
+    Returns:
+        bpy.types.Object: Top object
+    """
+    while obj.parent:
+        obj = obj.parent
+    return obj
```
MustafaJafar, line 344: Could you provide a test case for it? 
e.g. a screenshot of a hierarchy and the expected output when running that piece of code on it.

moonyuet: It should be working for empty? 
image

MustafaJafar:
Is this expected? It took only one object of my selection as opposed to taking the two objects before this PR.
- In this PR 
Animation_1058
- In current develop
Animation_1059

client/ayon_blender/plugins/create/create_model.py
```diff
@@ -24,8 +22,8 @@ def create(
 
         # Add selected objects to instance
         if pre_create_data.get("use_selection"):
-            bpy.context.view_layer.objects.active = asset_group
-            for obj in lib.get_selection():
-                obj.parent = asset_group
+            selections = lib.get_selection()
+            top_root = lib.get_highest_root(selections)
+            top_root.parent = asset_group
```
BigRoy, line 27: Please revert this change.

This should just be:
```python
for obj in selection:
    obj.parent = asset_group
```

The user selects, without hierarchy, what they want to add into the asset group. If they happen to select children as well, then that will also get parented to the asset group - that's absolutely fine behavior. They'll learn and see that if that's not what they want, then don't select the children - select just the root, etc.

client/ayon_blender/plugins/create/create_camera.py
```diff
@@ -26,8 +26,9 @@ def create(
 
         bpy.context.view_layer.objects.active = asset_group
         if pre_create_data.get("use_selection"):
-            for obj in lib.get_selection():
-                obj.parent = asset_group
+            selections = lib.get_selection()
+            top_root = lib.get_highest_root(selections)
+            top_root.parent = asset_group
```
BigRoy, line 31: Please revert this change.

This should just be:
```python
for obj in selection:
    obj.parent = asset_group
```

The user selects, without hierarchy, what they want to add into the asset group. If they happen to select children as well, then that will also get parented to the asset group - that's absolutely fine behavior. They'll learn and see that if that's not what they want, then don't select the children - select just the root, etc.

client/ayon_blender/api/lib.py
```diff
@@ -311,23 +311,44 @@ def get_selected_collections():
     return [id for id in ids if isinstance(id, bpy.types.Collection)]
 
 
-def get_selection(include_collections: bool = False) -> List[bpy.types.Object]:
+def get_selection(
+        include_collections: bool = False,
+        include_object_children_recursive: bool = True
```
BigRoy, line 316: Please default this to `False` - This is an odd default for a library function that is set out to return your current selection. It makes no sense from the perspective of the library function. If you want it enabled somewhere, then pass the argument and set it to True.

client/ayon_blender/plugins/load/load_camera_fbx.py
```diff
@@ -55,7 +55,8 @@ def _process(self, libpath, asset_group, group_name):
             obj.parent = asset_group
 
         for obj in objects:
-            parent.objects.link(obj)
+            if obj not in parent.objects:
+                parent.objects.link(obj)
```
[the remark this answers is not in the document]
BigRoy: Nevermind - it's to avoid `RuntimeError` that an object is already in the collection.

BigRoy:
After a call with @LiborBatek and @philippe-ynput it seems we may abandon this PR in favor of (optionally) collecting it during publishing as per PR https://github.com/ynput/ayon-blender/pull/134

Issue #129, opened 2025-07-03
Add `Create asset hierarchy` in `USD` creator.


MustafaJafar:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Add `Create asset hierarchy` in `USD` creator in compliance to `Maya USD` creator in Maya. 

### Suggested implementation?

.

### Describe alternatives you've considered:

.

PR #131, opened 2025-07-03
Enhancement/usd create extract enhancements

### LINKED ISSUE #130: `USD` export has default root prim `/root`

## Current Behavior
`USD` export has default root prim `/root`

## Expected Behavior
I expected to be the top level prim in my object hierarchy in the outliner as in Maya.

### Steps To Reproduce:

_No response_

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #130 #129 

- Use the top-level object as the root prim to match `Maya USD` UX.
- Add `Create asset hierarchy` toggle to the `USD` creator to match `Maya USD` UX.
- Expose orientation convention setting so that output matches the output of other DCCs.
image

## Some Test Runs
Step 2 in testing notes: 
image

<details><summary> example output file: Expr_suzanne_usdmodelMain_v002.usd
</summary>

```
#usda 1.0
(
    defaultPrim = "suzanne"
    doc = "Blender v4.2.2 LTS"
    metersPerUnit = 1
    upAxis = "Z"
)

def Xform "suzanne"
{
    custom string userProperties:blender:object_name = "suzanne"

    def Xform "geo"
    {
        custom string userProperties:blender:object_name = "geo"

        def Xform "Suzanne"
        {
            custom string userProperties:blender:object_name = "Suzanne"

            def Mesh "Suzanne_002" (
                active = true
            )
            {
                ...
                uniform token subdivisionScheme = "none"
                custom string userProperties:blender:data_name = "Suzanne.002"
            }
        }
    }
}
```

</details>

## Testing notes:
1. Using `USD` creator should be the as same as using `Maya USD` creator. steps [here](https://help.ayon.app/en/help/articles/7468381-ayon-maya-openusd-workflow#k0et1zsxtmf).
2. when examinign the output usd layer, you find the root prim is the same as the top-level object in the outliner.
3. Try setting orientation convention.

### DIFF
--- a/client/ayon_blender/plugins/load/load_blend.py
+++ b/client/ayon_blender/plugins/load/load_blend.py
@@ -1,4 +1,5 @@
-from typing import Dict, List, Optional
+from __future__ import annotations
+from typing import Optional
 from pathlib import Path
 
 import bpy
@@ -77,18 +78,23 @@ class BlendLoader(plugin.BlenderLoader):
 
     def _process_data(self, libpath, group_name):
         # Append all the data from the .blend file
+        names_by_attr: dict[str, list[str]] = {}
         with bpy.data.libraries.load(
             libpath, link=False, relative=False
         ) as (data_from, data_to):
             for attr in dir(data_to):
-                setattr(data_to, attr, getattr(data_from, attr))
-
-        members = []
+                values = getattr(data_from, attr)
+                # store copy of list of names because the main list will
+                # be replaced with the data from the library after the context
+                names_by_attr[attr] = list(values)
+                setattr(data_to, attr, values)
 
         # Rename the object to add the asset name
+        members = []
         for attr in dir(data_to):
-            for data in getattr(data_to, attr):
-                data.name = f"{group_name}:{data.name}"
+            from_names: list[str] = names_by_attr[attr]
+            for from_name, data in zip(from_names, getattr(data_to, attr)):
+                data.name = f"{group_name}:{from_name}"
                 members.append(data)
 
         container = self._get_asset_container(data_to.objects)
@@ -117,8 +123,8 @@ class BlendLoader(plugin.BlenderLoader):
 
     def process_asset(
         self, context: dict, name: str, namespace: Optional[str] = None,
-        options: Optional[Dict] = None
-    ) -> Optional[List]:
+        options: Optional[dict] = None
+    ) -> Optional[list]:
         """
         Arguments:
             name: Use pre-defined name
@@ -182,7 +188,7 @@ class BlendLoader(plugin.BlenderLoader):
         self[:] = objects
         return objects
 
-    def exec_update(self, container: Dict, context: Dict):
+    def exec_update(self, container: dict, context: dict):
         """
         Update the loaded asset.
         """
@@ -252,7 +258,7 @@ class BlendLoader(plugin.BlenderLoader):
             parent_container[AVALON_PROPERTY]["members"] = (
                 parent_members + members)
 
-    def exec_remove(self, container: Dict) -> bool:
+    def exec_remove(self, container: dict) -> bool:
         """
         Remove an existing container from a Blender scene.
         """
--- a/client/ayon_blender/plugins/load/load_blendscene.py
+++ b/client/ayon_blender/plugins/load/load_blendscene.py
@@ -1,4 +1,5 @@
-from typing import Dict, List, Optional
+from __future__ import annotations
+from typing import Optional
 from pathlib import Path
 
 import bpy
@@ -36,18 +37,23 @@ class BlendSceneLoader(plugin.BlenderLoader):
 
     def _process_data(self, libpath, group_name, product_type):
         # Append all the data from the .blend file
+        names_by_attr: dict[str, list[str]] = {}
         with bpy.data.libraries.load(
             libpath, link=False, relative=False
         ) as (data_from, data_to):
             for attr in dir(data_to):
-                setattr(data_to, attr, getattr(data_from, attr))
-
-        members = []
+                values = getattr(data_from, attr)
+                # store copy of list of names because the main list will
+                # be replaced with the data from the library after the context
+                names_by_attr[attr] = list(values)
+                setattr(data_to, attr, values)
 
         # Rename the object to add the asset name
+        members = []
         for attr in dir(data_to):
-            for data in getattr(data_to, attr):
-                data.name = f"{group_name}:{data.name}"
+            from_names: list[str] = names_by_attr[attr]
+            for from_name, data in zip(from_names, getattr(data_to, attr)):
+                data.name = f"{group_name}:{from_name}"
                 members.append(data)
 
         container = self._get_asset_container(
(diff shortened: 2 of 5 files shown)


client/ayon_blender/plugins/create/create_usd.py
```diff
@@ -20,11 +22,48 @@ def create(
...
+            # Parent members with geo and remove them from selection
             for obj in objects:
-                collection.objects.link(obj)
-                if obj.type == 'EMPTY':
-                    objects.extend(obj.children)
+                obj.parent = geo
+                obj.select_set(False)
+
+            # Select root
+            root.select_set(True)
```
BigRoy, line 47: Is the selection required for anything?

client/ayon_blender/plugins/create/create_usd.py
```diff
@@ -20,11 +22,48 @@ def create(
...
+            bpy.context.scene.collection.objects.link(geo)
+            geo.parent = root
+
+            # Parent members with geo and remove them from selection
             for obj in objects:
-                collection.objects.link(obj)
-                if obj.type == 'EMPTY':
-                    objects.extend(obj.children)
+                obj.parent = geo
+                obj.select_set(False)
```
BigRoy, line 44: Is the selection change required?

BigRoy (approved):
LGTM - didn't test but @LiborBatek could test it perhaps. Just not sure about the selection being applied - is that for ease of use? or is it required for the rest to work? seems a bit odd to do?

client/ayon_blender/plugins/create/create_usd.py
MustafaJafar: I don't like it anyways, removing it.

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -65,8 +76,13 @@ def process(self, instance):
                 # export_mesh_colors=True,
                 export_normals=True,
                 export_materials=True,
-                use_instancing=True
+                use_instancing=True,
+                # Convert Orientation
+                convert_orientation=convert_orientation,
+                export_global_forward_selection=forward_axis,
+                export_global_up_selection=up_axis,
```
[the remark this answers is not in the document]
BigRoy: I wonder if we should also add in the removal of the World Dome Light - with similar 'backwards compatibility'?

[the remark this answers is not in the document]
moonyuet: Their 4.5 is gonna be another LTS, which would be released after few days(9th july). https://[developer.blender.org/docs/release_notes/4.5/](https://developer.blender.org/docs/release_notes/4.5/)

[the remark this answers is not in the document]
MustafaJafar: I'm not against adding any backwards compatibility logic. I guess we may end up mentioning that 3.6 LTS is the minimum requried version but since it reached its EOL I wonder how many users are using it.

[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-blender/pull/131/commits/96efa2b8619b100a50d604c447708b788c512be8

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -50,11 +53,30 @@ def process(self, instance):
...
+            "export_global_forward_selection": attribute_values.get("forward_axis"),
+            "export_global_up_selection": attribute_values.get("up_axis"),
+        }
+        if lib.get_blender_version() < (4, 2, 1):
+            orientation_args = {}
+            if convert_orientation:
+                self.log.warning(
+                    "Convert orientation was enabled for USD export but is not "
+                    "supported in Blender < \"4.2.1\". Please update to at least Blender "
+                    "4.2.1 to support it"
```
BigRoy, line 70:
```suggestion
                    "4.2.1 to support it."
```

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -80,6 +104,75 @@ def process(self, instance):
...
+    
+    @classmethod
+    def get_attr_defs_for_instance(cls, create_context, instance):
+        # Filtering of instance, if needed, can be customized
+        if not cls.instance_matches_plugin_families(instance):
+            return []
+        
+        # Attributes logic
+        publish_attributes = instance["publish_attributes"].get(
+            cls.__name__, {})
```
BigRoy, line 116: I wonder if `self.get_attr_values_from_data` could be used here? @iLLiCiTiT thoughts?

client/ayon_blender/plugins/create/create_usd.py
```diff
@@ -20,11 +22,46 @@ def create(
...
+            geo = bpy.data.objects.new("geo", object_data=None)
+            bpy.context.scene.collection.objects.link(geo)
+            geo.parent = root
+
+            # Parent members with geo.
             for obj in objects:
-                collection.objects.link(obj)
-                if obj.type == 'EMPTY':
-                    objects.extend(obj.children)
+                obj.parent = geo
```
BigRoy, line 43: This would go quite bad with @moonyuet's changes being made here: https://github.com/ynput/ayon-blender/pull/128

Because then suddenly, all children under your current selection would then also get reparented to this root - which is not what we'd want. So we may need to handle this case separately @moonyuet @MustafaJafar

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -50,11 +53,30 @@ def process(self, instance):
...
+        
+        attribute_values = self.get_attr_values_from_data(instance.data)
+        convert_orientation = attribute_values.get("convert_orientation")
+        orientation_args = {
+            "convert_orientation": convert_orientation,
+            "export_global_forward_selection": attribute_values.get("forward_axis"),
+            "export_global_up_selection": attribute_values.get("up_axis"),
+        }
+        if lib.get_blender_version() < (4, 2, 1):
+            orientation_args = {}
```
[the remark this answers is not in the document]
MustafaJafar: > since future compatibility tweaks would be made in this same object.

Tbh, since we don't have any additional kwargs at the moment, I thought about it naming it a descriptive name like `orientation_kwargs`.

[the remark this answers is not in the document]
MustafaJafar: I guess I took it litteraly and used only `kwargs` 😅 
https://github.com/ynput/ayon-blender/pull/131/commits/008fcffd05eab5be1ab07ed686cb3c98349a8197

client/ayon_blender/plugins/create/create_usd.py
moonyuet: Only go quite bad when the condition of `checking the type of object is empty` is not removed once my PR is merged.

BigRoy: That won't solve it - a hierarchy can exist of multiple empties parented to each other, that hierarchy should be maintained.

BigRoy: With @moonyuet 's PR the easiest fix is to just NOT get the selection with hierarchy:

```python
selection = lib.get_selection()  # assuming this defaults to NOT including the unselected children hierarchy
for obj in selection:
   obj.parent = root

selection_hierarchy = set(selection) | lib.get_object_children_recursive(selection)
for obj in selection_hierarchy:
    collection.link(obj)
```

This is the most sensible approach here.

PR #133, opened 2025-07-03
Link Scene by Library

moonyuet:
## Changelog Description
This PR is to add some adjustment for linking the asset and imprinting the ayon container data into the libraries.

@LiborBatek feel free to investigate once you perform the test with this. 
- If you load multiple assets with the same version, the container only counts one data as the assets shared the same libraries. This also means you cannot link the multiple assets with the same version and versioning the assets innovatively as if we did with loading the asset per collection.
- There is a scenario when users load assets with different versions (asset A and B), try to up-version asset A to be the same version with Asset B, the updating version would be errored out because they have same library filepath shared (Some StructRNA errors)
- Linking system in Blender /= Referencing in Maya. The library datablock only allows to be loaded once per filepath. And once the data is unused, it would be removed automatically.
- The contributor has the reason behind to link the assets and imprint the container data into the collection instead. 

## Testing notes:
1. Load asset through link scene
2. Update or remove the assets

### DIFF
--- a/client/ayon_blender/plugins/load/load_blend_link.py
+++ b/client/ayon_blender/plugins/load/load_blend_link.py
@@ -1,10 +1,9 @@
 import bpy
-import os
 from typing import Dict, List, Optional, Union
 from ayon_core.lib import BoolDef
 from ayon_blender.api import plugin
 
-from ayon_core.pipeline import AVALON_CONTAINER_ID
+from ayon_core.pipeline import AYON_CONTAINER_ID
 from ayon_blender.api.plugin_load import (
     add_override,
     load_collection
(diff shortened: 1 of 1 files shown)


BigRoy:
I've taken a look and I think I'm even more confused than I was about Linking in Blender.
I'm so confused how to link e.g. the same library twice. Say I do that, I sometimes magically seem able to move them individually - but MOST of the time as I click "add override" or whatever it adds it to ALL of them - they seem literal instances (the same datablock)...

I don't think I'm the right person to look into this - or I need a quick tutorial on how this lining and library override actually works. I'm so lost. 

@LiborBatek say you link a collection twice, how do I add an override to just ONE of them? (You can't right?)

moonyuet:
> I don't think I'm the right person to look into this - or I need a quick tutorial on how this lining and library override actually works. I'm so lost.
> 
> @LiborBatek say you link a collection twice, how do I add an override to just ONE of them? (You can't right?) As soon as I do "add override" it does so to ALL instances (which makes sense) but then if I add override, and load a new one... then suddenly I have only one with an override but both from the same datablock? 🤯

I am also super confused about this too. To be honest, I really hope that they can introduce the library datablock similarly to how they did in the cache files, 
I would try to use hacky way to achieve this. (as you can see in the TODO list)

BigRoy:
If we're now allowing to load a library only once (or make 'untracked' instances of it) then essentially the approach to imprint on the library itself does make more sense again perhaps?

moonyuet:
> If we're now allowing to load a library only once (or make 'untracked' instances of it) then essentially the approach to imprint on the library itself does make more sense again perhaps?

Logically yes if we are dealing with just library. But the contributor has some reason behind to link the assets and imprint the container data into the collection instead(if you studied the code, the function containerise also imprinting the collection), while they just supported to link once.

BigRoy:
> Logically yes if we are dealing with just library. But the contributor has some reason behind to link the assets and imprint the container data into the collection instead(if you studied the code, the function containerise also imprinting the collection), while they just supported to link once.

I have a feeling that's purely because they were just re-using the "containerize" logic across the board - I suspect that's not intentional. Anyway, seems now again there _may_ actually be a use case again for loading libraries more than once... so 🤷‍♂️ wait and see.

PR #134, opened 2025-07-04
Include object child hierarchy when publishing Collection instances

BigRoy:
## Changelog Description

Collect full hierarchy of objects inside a collection as part of the instance, even if they are not directly linked as objects to the collection.

This way, if someone only adds the root group in the publish collection, the resulting publish will still include all children.

Fix https://github.com/ynput/ayon-blender/issues/127 but instead of changing what is added 'automagically' on create of the instance  (like [here](https://github.com/ynput/ayon-blender/pull/128)), the publishing behavior is just fixed to include all children objects.

## Testing notes:
1. create the instance again like this video below [source](https://github.com/ynput/ayon-blender/pull/128#issuecomment-3035548239)

https://github.com/user-attachments/assets/c9dc84e0-5974-497b-8d6b-36f47ed27d9e

2. publish blend scene should include the child objects


BigRoy:
So this now does the same as what @moonyuet tried to do in https://github.com/ynput/ayon-blender/pull/128 but without requiring the full hierarchical selection _inside_ the Collection.

### Pros

- No need to have more objects linked in the collection
- If you add a new object (e.g. Cube) to the hierarchy after Create it will still be included if it's parented under e.g. the root that's added in the collection. In https://github.com/ynput/ayon-blender/pull/128 it requires to then still add that to the Collection manually.

### Cons

- Currently no way to 'disable the "include all children" behavior - as such, you can't now export a blendscene with just the parent cube _without_ its children.

---

@moonyuet @LiborBatek thoughts?

moonyuet:
> If you add a new object (e.g. Cube) to the hierarchy after Create it will still be included if it's parented under e.g. the root that's added in the collection. In https://github.com/ynput/ayon-blender/pull/128 it requires to then still add that to the Collection manually.

I am not sure why you are overcomplicating these a lot when it can be only achieved by just editing `get_selection` function in library and force `get_selection`  to get the hierarchy instead. (We dont need to edit the collector and creator)
I can totally make it mandatory without `select hierarchies` option and ask Blender to select the hierarchy directly. This would do the same with the code you did above and with less code. And users can choose to export the transform of the object only if they just want the transform data of the objects but not the shape object. (I used the term shape so that it matches how it sounds in Maya).

There are differences between `select` and `select hierarchy` in blender, the former one just selects the top of the object, the other selects the top of the objects and its hierarchies. And what the use selection achieves is `select` but not `select hierarchy`. Blender is not Maya, it won't select your object's transform and shape all at once on some occasions (In the case of the video call, the object mis-selected due to being parented to the empty. When you select the locator, it would only select your object's transform but not your object's shape.

BigRoy:
After a call with @philippe-ynput and @LiborBatek it seems it may actually be preferred to add the object children at publish times - as opposed to during "Create" only. Which makes this PR favored over https://github.com/ynput/ayon-blender/pull/128

Hence, reopening this.

moonyuet:
I tried with a bigger scene(not really pretty one forgot me) I believe the selections work well and keep as expected(with or without select hierarchy, it still gets into being part of the AYON instances correctly without greyed out). But the grey out selection seems to happen occasionally.
I did some stress test(with geometry node, particle cloud and some huge textures with linked assets, some textures on the ground)
But I would like to suggest that it would be nice to have an option to filter linked objects and the packed texture (aka makes it unpacked) for blendScene publishing. 

https://github.com/user-attachments/assets/ba11e7d2-b166-4718-a331-2e9a5dbb8358

LiborBatek:
As I have been testing this PR....I have found out something to consider...

When making some selection of entities before creating ```blendScene``` publish instance type (with ```Use Selection``` enabled)

the resulting ```AVALON_INSTANCES``` collection (marked by green rectangle on the img below) contain just purely those preselected entities by user before triggering the creation of publish instance...

Now to my question...as you can also see we are missing all the hierarchical entities above those objects (marked red) , I really cant tell whether its good or bad but we should consider it for sure. As far as I know @BigRoy prefers to be really exact on user selection not altering it anyhow and just optionally include full hierarchy during publishing phase (we miss this option atm - couldnt test it yet myself).


Screenshot 2025-07-11 162245

So in essence the user will get just what he selected and will loose any Collections which the object sit in e.g. ```Box01.001``` wont be sitting in collection ```CloseBoxes``` and nested one ```MiddleBox```.

LiborBatek:
Anyway... need to be tested when ```Include children``` publishing option will be on place.

BigRoy:
Can you share a reproducable scene (starting state) and describe what to do to get the duplicated objects? 
What did you link/include in the Ayon instance?

PR #136, opened 2025-07-04
Append Blend: avoid objects renamed on name clash

### LINKED ISSUE #135: Load: Append Blend has renamed objects if object with that name already exists in the current scene

## Current Behavior

When doing "Append scene" and the loaded scene contains e.g. `Cube` and the current scene also has `Cube`.
Then on load the object becomes `namespace:Cube.001` or alike.

## Expected Behavior

Original name should be preserved, so that it is `namespace:Cube`.
The namespace should be what is making it unique.

### Steps To Reproduce:

1. Create Cube
2. Add it to blend scene product type publish.
3. Publish
4. In same scene, load the publish.

The loaded object should be named `namespace:Cube` but gets renamed to `Cube.001` before we append the `namespace:` to it.

### Additional context:

Blender 4.4.3
ayon-blender 0.2.11+dev

### Relevant log output:

```shell

```

BigRoy:
## Changelog Description

Avoid objects being renamed on load due to name clash before we're able to prepend with a namespace.

During the `bpy.data.libraries.load` context manager the datablock lists are lists of names, but after the context they are the actual objects that may have already been renamed on a name clash. So with this PR we keep a copy of list of the original names being loaded so we can use those names instead when we prepend the namespace from the group name.

Fix #135

## Testing notes:

1. Create Cube
2. Add it to blend scene product type publish.

image

4. Publish
5. In same scene, load the publish.

**Before PR:**
image

**After PR:**
image

The loaded Cube should be named `namespace:Cube` instead of `namespace:Cube.001`


client/ayon_blender/plugins/load/load_blend.py
```diff
@@ -117,15 +123,16 @@ def _process_data(self, libpath, group_name):
...
+        options: Optional[dict] = None
+    ) -> Optional[list]:
         """
         Arguments:
             name: Use pre-defined name
             namespace: Use pre-defined namespace
             context: Full parenthood of representation to load
             options: Additional settings dictionary
         """
+        print("HELLO WORLD")
```
moonyuet, line 135:
```suggestion

```

PR #137, opened 2025-07-04
Validate File Saved: Less strict + always save current scene on publish

BigRoy:
## Changelog Description

Be less strict to the user on workfile having no unsaved changes in validations.

- Now always save during early Extractor stage, to ensure no unsaved changes.
- Only validate whether file has never been saved (has no filepath) because we need a filepath
    - Skip this validation in Blender if similar implementation exists in AYON core.
- Remove `exclude_families` setting for ValidateFileSaved because according to Ynput statistics it's essentially unused, plus since this validation is now legacy or deprecated due to addition to ayon-core there's no need for that unique behavior to remain there.

The validator that's now marked legacy is to be replaced by: https://github.com/ynput/ayon-core/pull/1363

image

With that validator you can also directly save without needing to open workfiles tool.

#### Note during testing

If the validation does not pass the `ayon-core` version check (which it likely won't because current release is 1.4.0 and not newer than that yet) but it _does already have the validation there then it may happen that both validations will be present (so be aware of that, it's likely to happen during current testing at least.) 

image

## Testing notes:

1. Validation should kick in now only when current scene has never been saved.
2. It should be replaced by validation in ayon-core if ayon-core is of newer version.

### DIFF
--- a/client/ayon_blender/plugins/load/load_blend.py
+++ b/client/ayon_blender/plugins/load/load_blend.py
@@ -1,4 +1,5 @@
-from typing import Dict, List, Optional
+from __future__ import annotations
+from typing import Optional
 from pathlib import Path
 
 import bpy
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
Works exactly like described in the info for the PR... nice!

LGTM
Screenshot 2025-07-08 151039

Issue #132, opened 2025-07-03
Ensure no animation/action timeline to be exported for the rig


moonyuet:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

The current integration of Blender allows extra animation data/action timeline to be exported for the rig.
We need to ensure this won't be exported as part of the data as it may/would affect how the asset works when linking it back to different scenes with link scene loader

### Suggested implementation?

There should be the mandatory validator for the rig product to ensure no animation data exported along with the armature.

PR #139, opened 2025-07-08
Implementation of Validate No Animation

moonyuet:
## Changelog Description
This PR is to implement the optional validator to check if the objects have animation data for model, rig and blendScene product, similar to what we have to check no animation data for model product in maya.

Resolve https://github.com/ynput/ayon-blender/issues/132

Tested with the objects with or without animation data.

## Testing notes:
1. Create BlendScene, Rig, Model instance by selections
2. Publish

### DIFF
--- a/client/ayon_blender/plugins/publish/validate_no_animation.py
+++ b/client/ayon_blender/plugins/publish/validate_no_animation.py
@@ -4,12 +4,11 @@ import inspect
 from ayon_core.pipeline.publish import (
     ValidateContentsOrder,
     OptionalPyblishPluginMixin,
-    PublishValidationError
+    PublishValidationError,
+    RepairAction,
 )
-from ayon_blender.api.action import (
-    SelectInvalidAction,
-    RepairAction
-)
+from ayon_blender.api.action import SelectInvalidAction
+
 from ayon_blender.api import plugin
 
 
@@ -23,6 +22,7 @@ class ValidateNoAnimation(
     hosts = ["blender"]
     families = ["blendScene", "model", "rig"]
     label = "No Animation"
+    optional = True
     actions = [SelectInvalidAction, RepairAction]
 
     @classmethod
--- a/client/ayon_blender/plugins/publish/validate_no_animation.py
+++ b/client/ayon_blender/plugins/publish/validate_no_animation.py
@@ -1,12 +1,15 @@
 
 import bpy
-
+import inspect
 from ayon_core.pipeline.publish import (
     ValidateContentsOrder,
     OptionalPyblishPluginMixin,
     PublishValidationError
 )
-import ayon_blender.api.action
+from ayon_blender.api.action import (
+    SelectInvalidAction,
+    RepairAction
+)
 from ayon_blender.api import plugin
 
 
@@ -14,13 +17,13 @@ class ValidateNoAnimation(
     plugin.BlenderInstancePlugin,
     OptionalPyblishPluginMixin
 ):
-    """Ensure that meshes don't have animation data."""
+    """Ensure that meshes do not have animation data."""
 
     order = ValidateContentsOrder
     hosts = ["blender"]
     families = ["blendScene", "model", "rig"]
     label = "No Animation"
-    actions = [ayon_blender.api.action.SelectInvalidAction]
+    actions = [SelectInvalidAction, RepairAction]
 
     @classmethod
     def get_invalid(cls, instance):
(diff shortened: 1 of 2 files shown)


LiborBatek (commented):
Been tested with ```Rig``` ```BlendScene``` and ```Model```

working nicely, also tried to disable/enable via settings and all working well.

There is only one suggestion as only removing key frames from the entity is not enough and user must get rid of the ```Action``` datablock too....even if no animation key frames present. It makes sense but could be nice to have ```Repair``` action which could delete any ```Action``` datablocks from the influenced objects.

Screenshot 2025-07-08 152805


Screenshot 2025-07-08 153413


Screenshot 2025-07-08 154030

client/ayon_blender/plugins/publish/validate_no_animation.py
```diff
@@ -0,0 +1,45 @@
...
+        if not self.is_active(instance.data):
+            return
+
+        invalid = self.get_invalid(instance)
+        if invalid:
+            names = ", ".join(obj.name for obj in invalid)
+            raise PublishValidationError(
+                "Objects found in instance which have"
+                f" animation data: {names}",
+                title="Keyframes on Objects"
```
BigRoy, line 44: Can you implement a nice `description` for this validation error too? So supply a `description` argument with a nice markdown formatted wording what the validation is about, etc. similar to [this](https://github.com/ynput/ayon-core/blob/f2b67e3d79497b6f2c9b26fc33edd353a784235d/client/ayon_core/plugins/publish/validate_file_saved.py#L50-L60). That way the right side of Publisher UI will show a description about the validation itself.

client/ayon_blender/plugins/publish/validate_no_animation.py
```diff
@@ -0,0 +1,45 @@
...
+
+class ValidateNoAnimation(
+    plugin.BlenderInstancePlugin,
+    OptionalPyblishPluginMixin
+):
+    """Ensure that meshes don't have animation data."""
+
+    order = ValidateContentsOrder
+    hosts = ["blender"]
+    families = ["blendScene", "model", "rig"]
```
BigRoy, line 21: Why does a `blendScene` disallow containing animation? That seems odd? Isn't that a "Allow anything" product type? @LiborBatek ?

BigRoy: > I personally dont mind to have such validator for `blendScene` too...just allowing option for more control over content? And still optional too...not an issue imho

Do beware that the default state for it being enabled or active, etc. in settings is not configurable per product type currently - so you may want a model and rig to **require** having no animation (non-optional), but a blend scene (most likely allow it instead) since it's like an "allow me any data".

client/ayon_blender/plugins/publish/validate_no_animation.py
```diff
@@ -0,0 +1,64 @@
...
+
+            Objects must not contain any keyframe animation data.
+            Please remove all keyframes before publishing.
+        """)
+
+    @classmethod
+    def repair(cls, instance):
+        invalid_objects = cls.get_invalid(instance)
+        for obj in invalid_objects:
+            obj.animation_data.action = None
```
BigRoy, line 64: Out of curiosity - in Blender, if you do this. Does it actually _delete_ the action/animation datablock if after this it has no users of that datablock anymore. Or does it remain in the scene data?

BigRoy (approved):
Didn't test but LGTM - I have my doubts about having the validator also target `blendScene` however. And I wonder if the current `repair` method may leave unused datablocks in the scenefile, but doesn't sound all too problematic?

client/ayon_blender/plugins/publish/validate_no_animation.py
moonyuet: It will still last in datablock, it is just removed in the action editor. You can always re-link it back if you want to.
image

client/ayon_blender/plugins/publish/validate_no_animation.py
LiborBatek: > > I personally dont mind to have such validator for `blendScene` too...just allowing option for more control over content? And still optional too...not an issue imho
> 
> Do beware that the default state for it being enabled or active, etc. in settings is not configurable per product type currently - so you may want a model and rig to **require** having no animation (non-optional), but a blend scene (most likely allow it instead) since it's like an "allow me any data".

@BigRoy 
It is a fair point tbh...but lets keep that for later if someone will argue about it. But didnt think of this scenario which makes very sense (rigs without anim as mandatory by default and non optional vs blendScene having the validator Disabled and/or exposed to user to tweak to his liking)

PR #140, opened 2025-07-08
Use `get_attr_values_from_data` in `ExtractUSD`

MustafaJafar:
## Changelog Description
Follow up PR to #131 as mentioned in https://github.com/ynput/ayon-blender/pull/131#discussion_r2192548549

- Use `get_attr_values_from_data` infavor of accessing instance data.

## Testing notes:
1. everything should work as usual.

### DIFF
--- a/client/ayon_blender/plugins/publish/extract_usd.py
+++ b/client/ayon_blender/plugins/publish/extract_usd.py
@@ -112,8 +112,7 @@ class ExtractUSD(plugin.BlenderExtractor,
             return []
         
         # Attributes logic
-        publish_attributes = instance["publish_attributes"].get(
-            cls.__name__, {})
+        publish_attributes = cls.get_attr_values_from_data(cls, instance.data)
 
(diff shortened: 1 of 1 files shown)


client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -112,8 +112,7 @@ def get_attr_defs_for_instance(cls, create_context, instance):
             return []
         
         # Attributes logic
-        publish_attributes = instance["publish_attributes"].get(
-            cls.__name__, {})
+        publish_attributes = cls.get_attr_values_from_data(cls, instance.data)
```
iLLiCiTiT, line 115: I didn't realize we're working with create instance here. This will probably work, but in that case it is not necessary to use `.data`.
```suggestion
        publish_attributes = cls.get_attr_values_from_data(cls, instance)
```

MustafaJafar:
This is awesome. it works very well. 
Animation_1060

Issue #141, opened 2025-07-09
Removing and adjusting 'AVALON' collections from the blender workfile as its not used anymore


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Simply getting rid of unused entities ```AVALON_CONTAINERS``` from blender workfile makes so much sense to lessen the clutter in the outliner etc.

<img width="289" height="301" alt="Image" src="https://github.com/user-attachments/assets/706af100-a423-4881-9d50-496895f954c6" />

Also touch up the way ```AVALON_INSTANCE``` been shown can be benefical for the better/easier usage. I propose to set it as "Inactive" for greying it out within the Outliner and/or make it Ayon green colored so the user can clearly distinguish between his/Ayon collection...

e.g. like this?

<img width="235" height="300" alt="Image" src="https://github.com/user-attachments/assets/7c9c017f-eda9-4d12-85c6-e2b0ffa43750" />

### Suggested implementation?

See above.

### Describe alternatives you've considered:

None

moonyuet:
It is possible to hide `ayon_instance` and its hierarchies but we need to consider some scenario when we are creating instances with Model Creator. We need to make sure the object is active for selection. we can probably achieve by using contextlib but need to consider if it would slow down the publishing process for the complex scene.
The color_tag is totally fine

<img width="1611" height="860" alt="Image" src="https://github.com/user-attachments/assets/5df22461-dfe8-49db-8d02-83b6f167d89e" />

Looks good for `BlendScene`
<img width="306" height="310" alt="Image" src="https://github.com/user-attachments/assets/4371b681-7a42-4119-aa65-b6650bb31a7c" />

LiborBatek:
@moonyuet I get that ... somaybe just keeping it in ```active``` state might do too!

another question,
Do you think its possible to control the order in which the ```Collection``` been shown in the outliner list?

e.g. to force the order... lot of times ```AYON_INSTANCE``` appears somewhere in the middle! Preferably the very top aka as first entry would be uber cool!

And usually after triggering AYON>Load the Outliner got cluttered and super busy by rolling out full hierarchies it would be much more preferable to force collapse the list to root nodes/Collections instead... I will provide some scrn shot to illustrate that...

moonyuet:
[...]

PR #144, opened 2025-07-15
Use color tags and add fake users for AYON-related data 

moonyuet:
## Changelog Description
This PR is to use color tags and add fake users for ayon-related data. It also makes sure workfile product imprinting on AYON instances. 
This PR also fixes the issue on how the loaded assets get their unique numbers, they are no longer dependent on checking the children from `AYON_CONTAINERS` 

Fixes https://github.com/ynput/ayon-blender/issues/141

n/a

## Testing notes:
1. Launch Blender
2. Create instances
3. Publish with workfile product and all should be successfully published
4. Load any asset(blendScene, rig, model etc.) should be all working as expected.

### DIFF
--- a/client/ayon_blender/api/plugin.py
+++ b/client/ayon_blender/api/plugin.py
@@ -255,9 +255,8 @@ class BlenderCreator(Creator):
                 Those may affect how creator works.
         """
         # Get Instance Container or create it if it does not exist
-        ayon_instances = bpy.data.collections.get(AYON_INSTANCES)
-        if not ayon_instances:
-            ayon_instances = self._create_ayon_instances_collection()
+        ayon_instances = self._ensure_ayon_instances_collection()
+
         # Create asset group
         folder_name = instance_data["folderPath"].split("/")[-1]
 
@@ -392,12 +391,17 @@ class BlenderCreator(Creator):
             }
         )
 
-    def _create_ayon_instances_collection(self) -> bpy.types.Collection:
+    def _ensure_ayon_instances_collection(self) -> bpy.types.Collection:
         """Create AYON Instances collections that contains created instances.
 
         Returns:
             bpy.types.Collection: AYON Instances collection
         """
+        node = bpy.data.collections.get(AYON_INSTANCES)
+        if node:
+            # Already exists, return it
+            return node
+
         node = bpy.data.collections.new(AYON_INSTANCES)
         node.color_tag = "COLOR_04"
         node.use_fake_user = True
--- a/client/ayon_blender/plugins/create/create_workfile.py
+++ b/client/ayon_blender/plugins/create/create_workfile.py
@@ -118,7 +118,7 @@ class CreateWorkfile(BlenderCreator, AutoCreator):
             data = created_instance.data_to_store()
             node = created_instance.transient_data.get("instance_node")
             if not node:
-                node = self._create_ayon_instances_collection()
+                node = self._ensure_ayon_instances_collection()
             else:
                 node = self._transfer_workfile_property(node)
 
@@ -145,9 +145,7 @@ class CreateWorkfile(BlenderCreator, AutoCreator):
         ):
             return node
 
-        instance_node = bpy.data.collections.get(AYON_INSTANCES)
-        if not instance_node:
-            instance_node = self._create_ayon_instances_collection()
+        instance_node = self._ensure_ayon_instances_collection()
         instance_node[AYON_PROPERTY] = node.get(AYON_PROPERTY)
         del node[AYON_PROPERTY]
         return instance_node


client/ayon_blender/api/plugin.py
```diff
@@ -66,10 +63,13 @@ def get_unique_number(
...
         if c.get(AYON_PROPERTY)}
-    container_names = obj_group_names.union(coll_group_names)
+    container_names = {
+        name.lower() for name
+        in obj_group_names.union(coll_group_names)
+    }
     count = 1
     name = f"{folder_name}_{count:0>2}_{product_name}"
-    while name in container_names:
+    while name.lower() in container_names:
```
BigRoy, line 72: Why do we suddenly want case insensitive matching? Seems odd?

client/ayon_blender/plugins/create/create_workfile.py
```diff
@@ -86,15 +85,17 @@ def create(self):
...
+            instance_node.color_tag = "COLOR_04"
+            instance_node.use_fake_user = True
+            bpy.context.scene.collection.children.link(instance_node)
+
         workfile_instance.transient_data["instance_node"] = instance_node
 
     def collect_instances(self):
-
-        instance_node = bpy.data.collections.get(AYON_CONTAINERS)
+        instance_node = bpy.data.collections.get(AYON_INSTANCES)
```
BigRoy, line 98: This needs backwards compatibility? It should still be able to find the old data and move it to the new place perhaps?

If it can't find the new - try the old?

BigRoy (commented):
Seems to work - some notes however.

I also wonder whether it's good to disable the 'visibility' and 'render' state for this collection.

image

Perhaps even globally disabling it in viewports?
image

Although maybe that may appear more confusing to beginner Blender users?

image

client/ayon_blender/api/pipeline.py
```diff
@@ -492,9 +492,9 @@ def convert_avalon_containers():
         avalon_containers.name = AYON_CONTAINERS
 
 
-def add_to_ayon_container(container: bpy.types.Collection):
+def add_to_ayon_container(container: Union[bpy.types.Collection, bpy.types.Object]):
     """Add the container to the AYON container."""
-
+    convert_avalon_containers()
```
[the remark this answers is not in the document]
BigRoy: What I meant was:

```python
names = (
    AYON_CONTAINERS, 
    # Backwards compatibility
    AVALON_CONTAINERS 
)
for name in names:
    ayon_container = bpy.data.collections.get(name)
    if ayon_container:
        # Found existing container property
        break
else:
    # Generate new container
    ayon_container = ensure_ayon_container()
```

So basically:

1. Search for AYON_CONTAINERS
2. If not found, search for AVALON_CONTAINERS (backwards compatibility) -> **note that technically this should then do conversion?**
3. If not found, generate new AYON_CONTAINERS.

With the current logic existing AVALON_CONTAINERS will overwrite AYON_CONTAINERS if both happen to exist?

client/ayon_blender/api/plugin.py
```diff
@@ -204,6 +201,9 @@ def cache_instance_data(shared_data):
             cache_legacy = {}
             convert_avalon_instances()
             ayon_instances = bpy.data.collections.get(AYON_INSTANCES)
+            if ayon_instances:
+                ayon_instances.color_tag = "COLOR_04"
+                ayon_instances.use_fake_user = True
```
BigRoy, line 206: I don't think we should be updating existing collections - the issue with this is that it'll mean it'll always remain forced... if for whatever reason someone changed the color on purpose - this suddenly will revert it?

client/ayon_blender/api/plugin.py
```diff
@@ -254,6 +254,8 @@ def create(
         instances = bpy.data.collections.get(AYON_INSTANCES)
         if not instances:
             instances = bpy.data.collections.new(name=AYON_INSTANCES)
+            instances.color_tag = "COLOR_04"
+            instances.use_fake_user = True
```
BigRoy, line 258: The fake user is redundant here, right? because we link it to the scene collection? Or are we doing this just to ensure it's more persistent? or?

client/ayon_blender/plugins/create/create_workfile.py
```diff
@@ -112,9 +115,8 @@ def collect_instances(self):
...
-    def remove_instances(self, instances):
-        for instance in instances:
-            node = instance.transient_data["instance_node"]
-            del node[AYON_PROPERTY]
-
-            self._remove_instance_from_context(instance)
+    def _find_old_workfile_property(self):
+        instance_node = bpy.data.collections.get(AYON_CONTAINERS)
+        if instance_node:
+            return instance_node.get(AYON_PROPERTY)
```
BigRoy, line 121: This doesn't suffice, because in `collect` we still wouldn't find the `AYON_INSTANCES` collection on that _first_ collect in a legacy scene. We must consider this old one the actual instance node as well ... or somehow enforce moving the property over to the `AYON_INSTANCES` (and ensure it's existence).

client/ayon_blender/api/plugin.py
moonyuet: I am thinking if we should just make the color_tag or fake users for newly created ayon instances only

client/ayon_blender/plugins/create/create_workfile.py
moonyuet: If we dive into the develop branch and test with the workfile, and we should find out that the instance data for workfile is not imprinted into `AYON_CONTAINERS`. I am thinking if it is necessity for implementing the backward compatibility for containers as it does not really imprint the data at all.

BigRoy: It does, as soon as you save/publish once in publisher UI - it does not on just "reset" or open of the publisher UI however.

image

client/ayon_blender/api/pipeline.py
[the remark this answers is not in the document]
moonyuet: >With the current logic existing AVALON_CONTAINERS will overwrite AYON_CONTAINERS if both happen to exist?

The current logic we have is to rename `AVALON_CONTAINERS` to `AYON_CONTAINER` if any `AVALON_CONTAINERS` found....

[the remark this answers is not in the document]
BigRoy: That's what I mean, what if somehow I merge a legacy scene with a new scene. Both might be present? Is the renaming sufficient then or will it error? Sounds like we'd then have conflicting names because the new name already exists?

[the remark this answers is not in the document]
moonyuet: > That's what I mean, what if somehow I merge a legacy scene with a new scene. Both might be present? Is the renaming sufficient then or will it error? Sounds like we'd then have conflicting names because the new name already exists?

Sorry, my bad. I didn't think of that as this is some edge case. Maybe there are some users manually appending/linking collection of the scene with `AVALON CONTAINERS`.

client/ayon_blender/api/plugin.py
```diff
@@ -396,6 +391,44 @@ def set_instance_data(
...
+            or AYON_PROPERTY not in node
+        ):
+            return node
+
+        instance_node = bpy.data.collections.get(AYON_INSTANCES)
+        if not instance_node:
+            instance_node = self._create_instance_node()
+        instance_node[AYON_PROPERTY] = node.get(AYON_PROPERTY)
+        del node[AYON_PROPERTY]
+        return instance_node
```
BigRoy, line 418: This is Workfile Creator related - it should not be on the base `BlenderCreator` plug-in.

client/ayon_blender/api/plugin.py
```diff
@@ -54,9 +53,7 @@ def get_unique_number(
     folder_name: str, product_name: str
 ) -> str:
     """Return a unique number based on the folder name."""
-    ayon_container = bpy.data.collections.get(AYON_CONTAINERS)
-    if not ayon_container:
-        return "01"
+    ayon_container = bpy.context.scene.collection
```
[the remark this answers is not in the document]
moonyuet: Then it makes sense why we previously get all unique numbers depending on `AYON_CONTAINERS`. I will restore the code.

client/ayon_blender/api/plugin.py
```diff
@@ -54,9 +54,7 @@ def get_unique_number(
     folder_name: str, product_name: str
 ) -> str:
     """Return a unique number based on the folder name."""
-    ayon_container = bpy.data.collections.get(AYON_CONTAINERS)
-    if not ayon_container:
-        return "01"
+    ayon_container = bpy.data.collection.get(AYON_CONTAINERS)
```
BigRoy, line 57: Shouldn't this call `get_ayon_container`?

BigRoy (approved):
Fixed some minor things - now seems all good. @moonyuet

moonyuet:
> Fixed some minor things - now seems all good. @moonyuet Can you do final code review and merge if you're ok?

Double checked in my side. Looks good and no error encountered so far.

PR #148, opened 2025-07-22
Add description to Creators

BigRoy:
## Changelog Description

Add description to Creators

Improve description for each creator in publisher UI.

image

image

## Testing notes:

1. Descriptions should work, be correct, to the point and accurate.

### DIFF
--- a/client/ayon_blender/plugins/create/create_blendScene.py
+++ b/client/ayon_blender/plugins/create/create_blendScene.py
@@ -6,7 +6,7 @@ from ayon_blender.api import plugin, lib
 
 
 class CreateBlendScene(plugin.BlenderCreator):
-    """Generic .blend export writing linked datablocks"""
+    """Generic .blend export writing datablocks"""
 
     identifier = "io.ayon.creators.blender.blendscene"
     label = "Blender Scene"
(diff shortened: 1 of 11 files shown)


client/ayon_blender/plugins/create/create_model.py
```diff
@@ -6,10 +6,11 @@
 
 
 class CreateModel(plugin.BlenderCreator):
-    """Polygonal static geometry."""
+    """Polygonal static geometry"""
```
MustafaJafar, line 9: Should It mentions the supported representations? I guess blender doesn't have the same limitation as Houdini where we need to usd nodes instead of an API.

BigRoy: I would avoid getting that specific actually.

client/ayon_blender/plugins/create/create_blendScene.py
```diff
@@ -6,10 +6,11 @@
 
 
 class CreateBlendScene(plugin.BlenderCreator):
-    """Generic group of assets."""
+    """Generic .blend export writing linked datablocks"""
```
moonyuet, line 9: We can either link or append .blend from BlendScene product, But the docstring appears to indicate this product only exporting assets for linking scene only.
```suggestion
    """Generic .blend export writing datablocks"""
```

PR #151, opened 2025-07-28
Use removing instances inherited from blenderCreator

moonyuet:
## Changelog Description
This PR is to remove `removing_instances` from Create BlendScene and uses `removing_instances` from the Base BlenderCreator instead. It is to avoid the deletion of the published objects after removing instances.

n/a

## Testing notes:
1. Create BlendScene Instance with some selected objects
2. Removing instance
3.  All objects should still stay at the scene.

### DIFF
--- a/client/ayon_blender/api/plugin.py
+++ b/client/ayon_blender/api/plugin.py
@@ -348,16 +348,30 @@ class BlenderCreator(Creator):
             self.imprint(node, data)
 
     def remove_instances(self, instances: List[CreatedInstance]):
-
         for instance in instances:
             node = instance.transient_data["instance_node"]
-
             if isinstance(node, bpy.types.Collection):
-                for children in node.children_recursive:
-                    if isinstance(children, bpy.types.Collection):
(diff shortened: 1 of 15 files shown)


LiborBatek (changes requested):
It sort of works in some occasions but not in all circumstances.

Here I have just multiple objects put into Publish Instance Collection and this works when removing such publish instance from the Publisher UI:
Instance_Remove_A

But in the other hand not working (removing physical objects too) while having some hierarchy with empty in the publish Instance)
Instance_Remove_B

not sure if correct to have such hierarchy but it works/ worked before so I guess its valid scenario when speaking of Publish Instance removal and retaing all entities aka not altering the workfile anyhow.

moonyuet:
@LiborBatek As suggested by Copilot, there is a scenario we need to check by your scene to create, remove and then re-create instance again to see if the error is hit.
Can you check with the updated commit again? Thanks

antirotor (approved):
Did a quick and simple test and it works, code looks fine also.

PR #155, opened 2025-08-08
Add blender temporary scene render directory as clean up path/directory

moonyuet:
## Changelog Description
This PR is to add blender temporary scene render directory as clean up path/directory

Enable `ayon+settings://core/publish/CollectRenderedFiles/remove_files` so that it can also remove along with the rendered files.
We should think about if it is much better idea to add option to `force directory clean up`  in core addon instead. This collector can be broken under some scenario 
## Testing notes:
1. Publish Render
2. Check to see if the tmp folder has been cleaned up

### DIFF
--- a/client/ayon_blender/plugins/publish/collect_temporary_files_for_clean_up.py
+++ b/client/ayon_blender/plugins/publish/collect_temporary_files_for_clean_up.py
@@ -1,0 +1,25 @@
+import os
+import pyblish.api
+
+
+class CollectTemporaryFilesForCleanUp(pyblish.api.InstancePlugin):
+    """Compare rendered and expected files"""
+
+    label = "Collect Temporary Files for Clean Up"
+    order = pyblish.api.CollectorOrder - 0.1
+    targets = ["farm"]
+
+def process(self, instance):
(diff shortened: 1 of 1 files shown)


LiborBatek (commented):
I have set the cleanup option on

Screenshot 2025-08-11 121821

and it had successfully removed staging files from the work folder which is great

Screenshot 2025-08-11 120007

There are some leftovers tho (blender specific case) as we have to treat also the main render globals file output path and force it to subfolder ```../tmp``` as there is currently no way to use system temp folder. So these are still present even when cleaning is enabled.

 
Screenshot 2025-08-11 121754

and within the staging work /tmp folder:

Screenshot 2025-08-11 121716


So in general it works speaking of standard behaviour across pipeline but for ```blender``` host specifically it still do not clean all of the intermediate files...

Not sure if related or for other PR.... @moonyuet  let me know and I optionally approve it.

LiborBatek:
@moonyuet speaking of those ```tmp``` render files ...as we are in control where those been saved to, maybe we could use such information to remove/clean those afterwards?? as Ayon been responsible for the setting the temp output dir/filenames.

PR #162, opened 2025-08-26
Fix load user addons from paths when starting Blender.

rdelillo:
## Changelog Description
Fixes:
* Blender 4.4+, when using `AYON_BLENDER_USE_SYSTEM_PATH`, returned value was always `False`.
* When using `BLENDER_USER_SCRIPTS`, default user preferences were not loaded if `BLENDER_USER_SCRIPTS` was not explictely overrided.

## Testing notes:
image

Blender 4.4+
1. Ensure that when `AYON_BLENDER_USE_SYSTEM_PATH` is either set to `0` or `1`, this get properly detected in the starting code.

Blender<4.4
1. Ensure default user preferences are loaded enven if `BLENDER_USER_SCRIPTS` is not defined

### DIFF
--- a/client/ayon_blender/addon.py
+++ b/client/ayon_blender/addon.py
@@ -34,8 +34,9 @@ class BlenderAddon(AYONAddon, IHostAddon):
         # If `AYON_BLENDER_USE_SYSTEM_PATH` is set use `BLENDER_SYSTEM_PATH`
         # to initialize the Blender startup environment, otherwise use the
         # `BLENDER_USER_PATH`.
+        # The launch env is not applied yet, so we pass the env value
+        # from the launch environment
         use_system_path = env_value_to_bool(
-            env_key="AYON_BLENDER_USE_SYSTEM_PATH",
             value=env.get("AYON_BLENDER_USE_SYSTEM_PATH"),
         )
         if use_system_path:
--- a/client/ayon_blender/addon.py
+++ b/client/ayon_blender/addon.py
@@ -34,7 +34,10 @@ class BlenderAddon(AYONAddon, IHostAddon):
         # If `AYON_BLENDER_USE_SYSTEM_PATH` is set use `BLENDER_SYSTEM_PATH`
         # to initialize the Blender startup environment, otherwise use the
         # `BLENDER_USER_PATH`.
-        use_system_path = env_value_to_bool("AYON_BLENDER_USE_SYSTEM_PATH")
+        use_system_path = env_value_to_bool(
+            env_key="AYON_BLENDER_USE_SYSTEM_PATH",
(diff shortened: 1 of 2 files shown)


client/ayon_blender/api/lib.py
```diff
@@ -131,7 +131,11 @@ def new_paths():
 
 
 def append_user_scripts():
-    user_scripts = os.environ.get("AYON_BLENDER_USER_SCRIPTS")
+    default_user_prefs = os.path.join(
+        bpy.utils.resource_path('USER'),
+        "scripts",
+    )
+    user_scripts = os.environ.get("AYON_BLENDER_USER_SCRIPTS") or default_user_prefs
```
rdelillo, line 138: When some user preference path was explicitely defined through `BLENDER_USER_SCRIPTS` this used to be copied in `AYON_BLENDER_USER_SCRIPTS` and properly re-injected. But what happens when the user preference path was not explicitely defined ? We need to ensure the default Blender path is reported in this case.

client/ayon_blender/addon.py
```diff
@@ -34,7 +34,10 @@ def add_implementation_envs(self, env, _app):
         # If `AYON_BLENDER_USE_SYSTEM_PATH` is set use `BLENDER_SYSTEM_PATH`
         # to initialize the Blender startup environment, otherwise use the
         # `BLENDER_USER_PATH`.
-        use_system_path = env_value_to_bool("AYON_BLENDER_USE_SYSTEM_PATH")
+        use_system_path = env_value_to_bool(
+            env_key="AYON_BLENDER_USE_SYSTEM_PATH",
+            value=env.get("AYON_BLENDER_USE_SYSTEM_PATH"),
+        )
```
BigRoy, line 40: I think the `env_key` is redundant here if `value` is passed. Also, a comment may be nice here.
```suggestion
        # The launch env is not applied yet, so we pass the env value
        # from the launch environment
        use_system_path = env_value_to_bool(
            value=env.get("AYON_BLENDER_USE_SYSTEM_PATH"),
        )
```

client/ayon_blender/api/lib.py
BigRoy: I thought on Blender 4.4+ we use `BLENDER_SYSTEM_SCRIPTS` and hence we don't touch `BLENDER_USER_SCRIPTS` at all?

And in the older 4.3- case, `BLENDER_USER_SCRIPTS` points to the AYON paths and since it doesn't allow multiple paths we must also include any original `BLENDER_USER_SCRIPTS` value. However, at the point this code runs... Wouldn't it always have been set since we would have applied our custom AYON path to `BLENDER_USER_SCRIPTS` already?

Issue #46, opened 2024-09-18
Publishing Animation (.blend) produces empty file which is unusable 


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When using ```Animation``` publish instance type and enabled ```Extract blend``` option it produces ```.blend``` file which does not contain any data. (Been tested to manually append any data from published scene but nothing shown within the file)

### Expected Behavior:

Published ```.blend``` file should at least contain ```action``` data which then could be loaded and used later in production.

As now it does not have any usage being it empty.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Load some rig
2. Animate it
3. Create ```Animation``` instance for publish
4. Enable ```Extract Blend``` option
5. Check the published file via Blender by File>Append and navigate to the published file
6. There wont be any content for appending	

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

moonyuet:
I am trying to replicate the issue, and I believe we need to elaborate a bit more of how it works.
When the animation is exported, the animation data are stored as data block, but it doesn't mean the asset is exported.
When the animation is loaded and linked, this needs to be done in the existing scene, which is similar to unreal. You need to have rig so that you can link your animation. 
Steps to test:
1. Animate your loaded rig
2. Publish
3. New scene
4. Load your rig
5. Link your animation.

LiborBatek:
Thanks, I wouldnt mind if this publish type/product would work only on appended products first (as a current limitation)...and maybe introduce the functionality for linked libraries later on in different PR?? Just my though.

Opinions?

moonyuet:
We can at least make `link animation` works for the animation product but the general library linker can be in the separate PR.

PR #166, opened 2025-09-15
Publish and Load animation with the related datablocks

moonyuet:
## Changelog Description
This PR is to make sure the animation product would produce the relevant data blocks(actions and objects). It also supports to append scene for animation product so that the animation linker can link action back to the character

Fix https://github.com/ynput/ayon-blender/issues/46

n/a
## Testing notes:
1. Publish animation with animated character
2. Append workfile
3. Link animation


LiborBatek (commented):
Ok, so I have published ```Animation``` with both ```Abc``` and ```.blend``` repre without any issues and played with loading of those products...

All works but there some "bumps on the road" still. Let me cover those in short.

When inspecting the ```Loader``` and its actions:

Screenshot 2025-09-16 140027

I can use ```Append Blend (blend)``` to append ```Animation``` as a whole (rig with animation data) and works well and predictably.

The same been true for ```Link Blend (blend)``` and again working nicely and results into linked library containing animated character rig from ```Animation``` product.

Then the  ```Link Animation (blend)``` caused me some headache as I was expecting it will be acting very much the same as ```Link Blend (blend)``` but not. As its actually loading ```Action``` hodling just animation...which is fine and maybe even better but the wording is incorrect for the loading action also it is not linking but append action (as there are no any linked libraries after using it) again it doesnt mean failure in the end but the wording of the loader is off.

So imho the ```Link Animation (blend)``` should be actually named e.g. ```Append Action (blend)``` or ```Append Animation (action)``` ...not really sure what are the rules for the labels within loader so I will leave it to you.

Some notes about above loader action (Link Animation (blend)) what it does it just brings in ```Action``` datablock without any ```Fake User``` also without noticing the user whats going on (at first I thought the loader failed for some reason as there is no visible change in the workfile as just the ```Action``` datablock got in. So it would also make sense to add some popup message telling user that he actually "just" appended ```Action``` datablock succesfully or similar...

@moonyuet what do you think?

Issue #161, opened 2025-08-26
`Append` loader action on `Workfile` product does not bring anything into blender scene but `Import` yes


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
There are actually 2 issues atm:

First issue:

When using ayon Loader and ```Append``` load on published ```Workfile``` product e.g. shot Animation workfile or Lighting workfile.

It wont bring anything into active blender scene.

<img width="563" height="398" alt="Image" src="https://github.com/user-attachments/assets/f5c0fbfe-bb3d-482e-a9c0-f12393f07975" />

Second issue:

Its weird that there are 2 loaders available ```Append load``` and ```Import``` as there is no such load possible in blender terminology as ```File > Import``` only works for non blender scenes.

<img width="377" height="476" alt="Image" src="https://github.com/user-attachments/assets/83ce4499-50d3-4375-9a58-894cc7c93c14" />

So we should remove/rename such ```Import``` load action  to ```Append``` load in fact. Resulting to only one available load type.

## Expected Behavior
Append loader should bring data into active scene as when performing it via native ```File>Append``` action in blender. And second, offer only `Append` action to the user as `Import` works only for non blender data types which obviously not been the case for ```Workfiles```

### Steps To Reproduce:

1. Open Ayon>Load
2. Select published `Workfile` of any kind and perform `Append` load
3. Nothing been loaded resulting into empty scene
4. Optionally test using `Import` load instead which actually brings stuff in

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #173, opened 2025-09-26
Remove import workfile loader to avoid confusion of `Append Workfile` and `Import Workfile`

moonyuet:
## Changelog Description
This PR is to remove import-workfile loader to avoid confusion of `Append Workfile` and `Import Workfile`
Resolve https://github.com/ynput/ayon-blender/issues/161

n/a

## Testing notes:
1. Load Workfile
2. Should only see append workfile


LiborBatek (approved):
Works within the scope of this PR (issues mentioned not relevant to this PR)

Approval then.

Issue #64, opened 2024-10-29
Publisher window got minimized during publishing and not visible again due to 'Review' publishing


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When doing publishing of ```Review``` and other products publisher window UI dissappears (got behind blenders ui) and not showing once publishing been finnished (stays at the background)

### Expected Behavior:

Publisher UI should stay visible or at least once the publishing done so user can read the outcome report easily. Also this behaviour been the standard for other DCC hosts like maya etc.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Create some product publish (model, rig etc)
2. Create 'Review' publish instance
3. Perform publishing
4. Publisher will dissappear behind host app UI and never get back on screen

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

BigRoy:
@LiborBatek does it minimize? or does it move to the background behind the currently running process?

Can you share a screen recording?

LiborBatek:
@BigRoy before I record the vid...yes, it is minimized to the windows tray...and stays there even the process been finnished

BigRoy:
> before I record the vid...yes, it is minimized to the windows tray...and stays there even the process been finnished

Thanks - it sounds very similar to something @Lypsolon mentioned happened for him - but instead of minimizing the window would actually 'accidentally' go behind the other active windows.

So, if you alt tab - can you still alt tab to the publisher UI and have it appear up front? if so, then maybe it wasn't mimized? but just on the background?

Lypsolon:
> Thanks - it sounds very similar to something @Lypsolon mentioned happened for him - but instead of minimizing the window would actually 'accidentally' go behind the other active windows.

my issue was related to the thumbnail generation. 
the generation will bring Houdini in the foreground and then take a "screen shoot" something like this i think and that moved Houdini in-front.

BigRoy:
> > Thanks - it sounds very similar to something @Lypsolon mentioned happened for him - but instead of minimizing the window would actually 'accidentally' go behind the other active windows.
> 
> my issue was related to the thumbnail generation. the generation will bring Houdini in the foreground and then take a "screen shoot" something like this i think and that moved Houdini in-front.

Ah corret - that is then Houdini specific. Ignore me 🙈

LiborBatek:
[...]

PR #180, opened 2025-10-16
Publisher UI always stays on top after review publishing

moonyuet:
## Changelog Description
This PR is to make sure Publisher UI is always top-level widget before-after publishing review.
Resolve https://github.com/ynput/ayon-blender/issues/64

n/a
## Testing notes:
1. Create Review
2. Publish
3. After finishing to extract review with the review screen closed, the publisher UI would stay at top again.

### DIFF
--- a/client/ayon_blender/plugins/publish/extract_playblast.py
+++ b/client/ayon_blender/plugins/publish/extract_playblast.py
@@ -1,6 +1,5 @@
 import os
 import json
-import contextlib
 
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Works like charm! I have tested it in multiple occasions with / without `Review` and all works nicely!

Simply stays on top!

Screenshot 2025-10-16 165140

Issue #189, opened 2025-10-24
Link Loading 'blendScene' product brings also Creator(s) into the active workfile which is not desirable


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
Whenever user load via `Link Blend` loader it also brings the `Creator` from the source scene meaning publisher got populated with publish instance which is not desirable as it causes unwanted publishes into original product context creating new version of the product itself.

## Expected Behavior
No publish instance should appear when link loading the `blendScene` products.
Or at least as in disabled state not potentially creating any publishes even when shown in the Creators list.

### Steps To Reproduce:

1. load existing `blendScene` product into active workfile
2. Inspect Publisher for Instances
3. Publisher/Publish tab lists unwanted `Publish Instance` originating from the `blendScene` product

### Version

`ayon-blender 1.0.8`

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #192, opened 2025-10-27
Deactivate the instance data during context for blendScene and blend animation publish

moonyuet:
## Changelog Description
This PR is to ensure the instance data is inactive during context for blendScene and blend animation publish. 
So that the relevant instance data won't be shown up in the Publisher UI once it is loaded back to blender via Ayon Loader.
Resolve https://github.com/ynput/ayon-blender/issues/189

n/a 

## Testing notes:
1. Create BlendScene
2. Publish
3. Load/Link back the same asset via ayon loader
4. Open the Publisher
5. You won't see the related data show up again.

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -617,15 +617,13 @@ def strip_container_data(containers):
 def strip_instance_data(node):
     """Remove instance data during context
     """
-    previous_data = node.pop(AYON_PROPERTY, None)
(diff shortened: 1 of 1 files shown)


LiborBatek:
@moonyuet I have tried also with different asset `blendScene` type and again still an issue:

Screenshot 2025-10-27 120619

note just not the publish instance but also the orig context which it is comming from (marked with arrow)

Issue #191, opened 2025-10-24
Render setup/ file output node for multilayered EXR been using lower case chars for `beauty` and `composite` causing issues when transcoded during publish DL job


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
Due to nature of blender AOVs and its names using capital letters we should use also capitals for render setup/file output node as its causing issues with layers order and publishing job which produces wrong reviewables then (using the very first layer using capital letters).

<img width="987" height="603" alt="Image" src="https://github.com/user-attachments/assets/c8c8f5db-00b2-43b7-b407-ffc56625e3f7" />

and when inspected the resulting EXR layers (note the order of the layers by naming leading to wrong layer `DiffCol` to be used for transcoding to mp4 using the very first layer...while `beauty` layer been moved by naming to the bottom)

Note: maybe we should also consider to rename the main file from `Main_beauty.###` to just `Main.####` while using `Multilayer EXR` output (located in the top area of the file output node on the screenshot.

<img width="1443" height="865" alt="Image" src="https://github.com/user-attachments/assets/1a8dde6b-2ef7-4907-95c1-b8a63a4c1344" />

## Expected Behavior
All layers should use capital letter as first character so they match the same naming pattern.

When manually adjusted to capitals before Render publishing all behaves correctly:

<img width="1445" height="866" alt="Image" src="https://github.com/user-attachments/assets/552ed7ee-7f51-44cc-b892-2986b84c23f6" />

### Steps To Reproduce:

1. Configure `ayon-blender` render settings to `Multilayer EXR`
2. Enable  additional AOVs like `DiffCol` which will take precedence over `beauty`
3. Publish Render and inspect the resulting rendered file layers and its Reviewable mp4 which uses wrong layer due to ordering by capitals

### Version
`ayon-blender 1.0.8`

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

moonyuet:
We need to add the check on the validator for the naming while we need to ensure the creator would create both beauty and composite name with capital letters

PR #193, opened 2025-10-27
AOVs: The 1st character capitalization for Beauty and Composite AOVs in compositor render output node

moonyuet:
## Changelog Description
This PR is to capitalize the first character for beauty and composite AOVs to avoid the transcoding issue during publish job submission to AYON.
Fixes: https://github.com/ynput/ayon-blender/issues/191

Putting this into the draft as I am still researching if it is possible to implement validator to fix this as there are two related slot nodes (file_slots and layer_slots) for this and layer slots is readonly. Working in progress to see if the other one is working. 
See: https://docs.blender.org/api/current/bpy.types.CompositorNodeOutputFile.html

## Testing notes:
1. Recreating your render instance
2. Publish

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -614,6 +614,18 @@ def strip_container_data(containers):
 
 
 @contextlib.contextmanager
+def strip_instance_data(node):
+    """Remove instance data during context
+    """
+    previous_data = dict(node.get(AYON_PROPERTY, {}))
+    try:
+        node[AYON_PROPERTY]["active"] = False
+        yield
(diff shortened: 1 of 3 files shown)


LiborBatek:
Now all works as expected...

Screenshot 2025-10-28 103538

Note: not sure, probably unrelated but atm my publish job fails on encoding via ffmpeg ...didnt happen until now, not sure if it is connected to naming changes even tho I have already tweaked my ayon settings for transcoding and reviewables altering those to capitals... but maybe that `Main.####.exr` ?? will investigate further tho

PR #196, opened 2025-10-29
Short labels for some plugins

### LINKED ISSUE #195: Some plugins names are too long which clutter the publisher UI

## Current Behavior
reported here internal discussion
https://discord.com/channels/517362899170230292/773127690453516298/1433014317912952952

## Expected Behavior
A concise description of what you expected to happen.

### Steps To Reproduce:

_No response_

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
Shorten labels for the following plugins and add tool tip:
- `ValidateCompositorNodeFileOutputPaths`
- `ValidateSceneRenderFilePath`

## Testing notes:
1. start with this step
2. follow this step

Resolve #195

### DIFF
--- a/client/ayon_blender/plugins/publish/validate_render_output_paths.py
+++ b/client/ayon_blender/plugins/publish/validate_render_output_paths.py
@@ -92,7 +92,10 @@ class ValidateSceneRenderFilePath(
     order = ValidateContentsOrder
     families = ["render"]
     hosts = ["blender"]
-    label = "Validate Scene Render Filepath"
+    label = "Validate Scene Output"
+    optional_tooltip = (
+        "Validate whether the scene-wide render filepath is set as required."
+    )
     optional = True
     actions = [RepairAction]
 


LiborBatek:
We could also alter that `Validate Deadline JobInfo` to e.g. `Validate Deadline Job` imho

BigRoy:
> We could also alter that Validate Deadline JobInfo to e.g. Validate Deadline Job imho

Admittedly, time is better spent somehow making optional validator plug-ins best listed together under a "Optional Validators" section and the labels having the `Validate ` prefix removed. The fact they are all prefixed with "Validate " feels so redundant.

LiborBatek:
> > We could also alter that Validate Deadline JobInfo to e.g. Validate Deadline Job imho
> 
> Admittedly, time is better spent somehow making optional validator plug-ins best listed together under a "Optional Validators" section and the labels having the `Validate ` prefix removed. The fact they are all prefixed with "Validate " feels so redundant.

so maybe new PR for that later on? once this one been merged? just to not prolong it and have quick fix atm?

BigRoy:
> so maybe new PR for that later on? once this one been merged? just to not prolong it and have quick fix atm?

It would be another PR regardless, because that one would be in `ayon-deadline` in particular. But I'd definitely raise the discussion with Philippe again about what we can do to streamline the UI experience for these attributes, etc.

For now @MustafaJafar just shorten the Validate Scene Render Filepath to e.g. `Validate Scene Output` and perhaps add a tooltip like: `optional_tooltip = "Validate whether the scene-wide render filepath is set as required."`

Then I'd say we can merge this.

Issue #97, opened 2025-02-06
Trying to load a missing image locks gui with a popup that cannot be reached


alainxi:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
If you try to use the Load gui to loadan image into Blender, and that image does not exist on disk anymore, the Load gui will freeze.

## Expected Behavior
The popup should appear in front of all others, not behind.

### Steps To Reproduce:

1 : From anywhere (for example Substance Painter), publish some image or texture.
2 : On disk, manually delete that image (but do not delete it from the database), as a tired artist would do.
3 : From inside Blender, use the menu Ayon-->Load. 
4 : Select the image you just published, and click on RightMouseButton-->Load in Compositor  (or Load to Shader Editor, etc...)

Blender and the Load gui will immediately freeze, and you will not be able to move or minimize the window anymore.

Using Alt-Tab, you will see there is a popup that cannot be raised to front (because the Load gui seems to always be the front-most, whatever you do), and so you cannot click on the popup's "Ok" button.

![Image](https://github.com/user-attachments/assets/8bc1800a-df4e-4753-bcb7-5c922870f798)

It can be reproduced by an artist.
Maybe it could be related to this other bug (which needs a developer to script stuff) :   https://github.com/ynput/ayon-blender/issues/65

### Additional context:

Windows 10.
Blender 4.0.2

### Relevant log output:

```shell
Traceback (most recent call last):

  File "C:\Users\Alain\AppData\Local\Ynput\AYON\addons\blender_0.2.8\ayon_blender\api\ops.py", line 129, in execute
    result = callback(*args, **kwargs)

  File "C:\Users\Alain\AppData\Local\Ynput\AYON\addons\blender_0.2.8\ayon_blender\api\plugin.py", line 483, in _load
    assert Path(filepath).exists(), f"{filepath} doesn't exist."

AssertionError: P:/UNREAL_POC_PRODUCTS/PROD/CREA/ASSETS_3D/TESTS/STAR/Surfacing/PUBLISH/textureSurfacingMain.base/v007/STAR_textureSurfacingMain.base_v007.png doesn't exist.
```

alainxi:
After having updated today (november 5th 2025) to the latest Ayon versions of everything (server, addons, dependencies), I have tested this bug again.
It is **not** fixed, the bug is still exactly the same.
[...]

Issue #65, opened 2024-10-29
QtWidgets appears behind other windows, and locks current window


alainxi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

From a Blender load addon, I want to display a Qt popup in front of Ayon’s Load gui
So, I tried this :

`from qtpy import QtWidgets
choice = QtWidgets.QMessageBox.warning(None,
                                       "The title",
                                        "The message",
                                        buttons = QtWidgets.QMessageBox.Ok,
                                        defaultButton = QtWidgets.QMessageBox.Ok)`

**Problem** : the popup appears behind the Load gui (which is obviously frozen because it waits for the user to click on the popup).
And you can’t Alt-Tab to switch to that pop-up (you still get the Load gui).

The good news : the popup displays the correct information (I can see it in the minimized windows that appear as part of the Alt-Tab process).

### Expected Behavior:

Any QtWidgets (volontary ones, or error popups) should display on top of the current window.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

It bugs today in Ayon with Blender 4.0.2 / PySide 6.

It already bugged a year ago in OpenPype with Blender 3.3.2 / PySide 2 :
https://github.com/ynput/OpenPype/issues/6016
https://github.com/ynput/OpenPype/issues/4304

That’s when I try to **display a popup on purpose.**

Now, let’s try to **display a popup as the result of an error.**
So, I take an Ayon addon for Blender (for example a Load addon), and I add any obvious python syntax error (not even related to Qt).
A standard Qt popup that displays the error will appear behind the current Ayon gui, and you can’t reach that popup.

**I have found a hack to make it work in Blender :**
`from ayon_core.tools.utils import host_tools
loaderGui_window = host_tools.get_tool_by_name("loader")
choice = QtWidgets.QMessageBox.warning(loaderGui_window, ............`

(I have replaced None by loaderGui_window)

Now the popup appears in front of the Ayon–>Load gui (and not behind anymore).
But it only works for a popup I code in my own addon, it obviously doesn’t work for popups that result of an error (as decribed earlier).

And it’s strange I have to specify the parent (other DCCs Load addons have the same kind of QtWidgets.QMessageBox that work all right with None as the parent).

Also, I’m not sure calling get_tool_by_name (to get the parent QtWidget) is a clean method.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[...]

PR #197, opened 2025-11-06
Make sure the error popup dialog would stay on top

moonyuet:
## Changelog Description
This PR is to fix the issue when users load something from loader and the loader raise exceptions, and the popup dialog does not show on top of the screen, and this freezes both blender and ayon plugins.
Resolve

Fix https://github.com/ynput/ayon-blender/issues/97
Fix https://github.com/ynput/ayon-blender/issues/65

## Testing notes:
1. Load something which can hit the error
2. The error popup should be always on top.

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -197,14 +197,9 @@ def _process_app_events() -> Optional[float]:
                 QtCore.Qt.WindowStaysOnTopHint |
                 QtCore.Qt.Dialog
             )
-
-            try:
-                dialog.show()
-                dialog.raise_()
-                dialog.activateWindow()
-                dialog.exec_()
-            finally:
-                dialog.deleteLater()
+            dialog.raise_()
+            dialog.activateWindow()
+            dialog.open()
 
         # Refresh Manager
         if GlobalClass.app:


client/ayon_blender/api/ops.py
```diff
@@ -188,7 +188,12 @@ def _process_app_events() -> Optional[float]:
                 msg)
             dialog.setMinimumWidth(500)
             dialog.setDetailedText(detail)
-            dialog.exec_()
+            dialog.setAttribute(QtCore.Qt.WA_DeleteOnClose)
```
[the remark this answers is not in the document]
BigRoy: Ah, interesting - then in this case we should be applying stylesheet here on the dialog too so that it looks correctly?

client/ayon_blender/api/ops.py
```diff
@@ -188,7 +189,22 @@ def _process_app_events() -> Optional[float]:
...
+                dialog.windowFlags() |
+                QtCore.Qt.WindowStaysOnTopHint |
+                QtCore.Qt.Dialog
+            )
+
+            try:
+                dialog.show()
+                dialog.raise_()
+                dialog.activateWindow()
+                dialog.exec_()
```
BigRoy: @iLLiCiTiT this might be a Qt question - but in this case we want to show the modal dialog to the user - but don't need to respond on their action. It's just an "error occurred" message.

According to Qt docs, `exec()` is to be avoided [like stated here](https://doc.qt.io/qt-6/qdialog.html#exec) and instead we should use `dialog.open()`. Is this a case where `.open()` does make more sense then? (I've never used it myself.)

iLLiCiTiT: Yep, `exec_` technically blocks the process, `open` just shows the message.

client/ayon_blender/api/ops.py
```diff
@@ -188,7 +189,22 @@ def _process_app_events() -> Optional[float]:
...
+            dialog.setAttribute(QtCore.Qt.WA_DeleteOnClose)
+            dialog.setStyleSheet(load_stylesheet())
+            # Ensure the dialog stays on top and is properly focused
+            dialog.setWindowFlags(
+                dialog.windowFlags() |
+                QtCore.Qt.WindowStaysOnTopHint |
+                QtCore.Qt.Dialog
+            )
+
+            try:
```
iLLiCiTiT: Why this is in try block with `deleteLater`?
1. Not sure what could fail?
2. If it fails it is cleaned up from memory and python+shiboken will take care of it.

moonyuet: If shiboken would take care of it, I would remove this then.

iLLiCiTiT: Also with using `open`, you don't want to call `deleteLater()` as it would destroy the window before it is closed 🙂.

iLLiCiTiT (approved):

Issue #126, opened 2025-07-02
Extract Blend: Packs ShadeNodeTexImage images, but should not change state in workfile


BigRoy:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior

When publishing anything that generates "Blend" scene file it packs all the ShadeNodeTexImage images from material slots.

https://github.com/ynput/ayon-blender/blob/ad6073a3228624d06e2ea49dcc4d1855e1b7d8ef/client/ayon_blender/plugins/publish/extract_blend.py#L46-L66 

Looking at the code - this will actually change the current scene and will make them packed... also in the current workfile - it doesn't "revert" that after.

## Expected Behavior

Workfile should remain unaltered.

### Steps To Reproduce:

1. Publish a workfile to `.blend` output with shader with images on geometry that are unpacked.
2. After publish, in current workfile they are most likely suddenly packed.

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #198, opened 2025-11-13
Extract Blend: Packed Image happened only during context 

moonyuet:
## Changelog Description
This PR is to make sure the current image pack only happened during context for extract blend, as discussed in https://github.com/ynput/ayon-blender/issues/126
Resolve https://github.com/ynput/ayon-blender/issues/126

Test with action product type to see if any issues hit. And also check the published blendScene and see if datablocks are all correctly located.

## Testing notes:
1. Create BlendScene or any product type(model, rig, action etc) 
2. Publish
3. Load the published asset back to blender and see if all the things are located in the datablocks menu.

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -665,7 +665,7 @@ def strip_namespace(containers):
 
 
 @contextlib.contextmanager
-def unpacked_images(datablocks):
+def packed_images(datablocks):
     """Unpack packed images during context
     This will pack all unpacked images found in the given datablocks,
     and unpack them back when exiting the context.
@@ -746,4 +746,3 @@ def search_replace_render_paths(src: str, dest: str) -> bool:
             changes = True
 
     return changes
-
--- a/client/ayon_blender/plugins/publish/extract_blend.py
+++ b/client/ayon_blender/plugins/publish/extract_blend.py
@@ -11,7 +11,7 @@ from ayon_blender.api.lib import (
     strip_container_data,
     strip_instance_data,
     strip_namespace,
-    unpacked_images,
+    packed_images,
 )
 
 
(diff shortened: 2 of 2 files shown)


client/ayon_blender/api/lib.py
```diff
@@ -664,6 +664,45 @@ def strip_namespace(containers):
...
+                if node.image and node.image.packed_file is None:
+                    unpacked_node_images.add(node.image)
+                    node.image.pack()
+    try:
+        yield
+
+    finally:
+        if unpacked_node_images:
+            for image in unpacked_node_images:
+                image.unpack()
```
BigRoy: If statement is redundant here. We'd just be iterating an empty set so it'd already skip it.
```suggestion
        for image in unpacked_node_images:
            image.unpack()
```

client/ayon_blender/plugins/publish/extract_blend.py
```diff
@@ -115,31 +117,9 @@ def add_datablock(self, instance: pyblish.api.Instance) -> set:
...
-                        continue
-                    # Check if image is not packed already
-                    # and pack it if not.
-                    if node.image and node.image.packed_file is None:
-                        node.image.pack()
-        return data_blocks
-
+        return {
+            data for data in instance
+        }
```
BigRoy: I think we can now remove this method and just do `data_blocks = set(instance)` above?

LiborBatek (approved):
I have tested all three `blendScene` `model` and `action` and all seems fine.

Speaking of products with `textures` all behave as expected and external files kept as external within the active `workfile` after publishing while when loaded via `Ayon>Load` its all correctly packed in.

When publishing (and after this):

Screenshot 2025-11-13 125451


While when loaded all got packed in:

Screenshot 2025-11-13 125837

PR #199, opened 2025-11-13
Image Loader: Make sure check_existing argument enabled for bpy.data.images.load

moonyuet:
## Changelog Description
This PR is to include the additional argument of `check_existing` into `bpy.data.images.load` to ensure image loader would always check if there was an existing image before load.

This fixes another potential error shown in the pop-up dialog of https://github.com/ynput/ayon-blender/issues/97
image

n/a

## Testing notes:
1. Launch Blender
2. Load some image with compositor and shader loaders

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -664,6 +664,44 @@ def strip_namespace(containers):
             node.name = f"{original_namespace}:{name}"
 
 
+@contextlib.contextmanager
+def packed_images(datablocks):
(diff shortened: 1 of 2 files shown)


BigRoy:
> This fixes another potential error shown in the pop-up dialog of #97

This change shouldn't fix that issue though? `check_existing` just makes it so that if you load it again, that it re-uses the existing datablock.

The change should only make it so that if you load the same image twice - that it uses the same existing datablock in the scene file.

Issue #203, opened 2025-11-28
YN-0246: Blender 5.0 compatibility


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Blender 5.0 changed a lot in their API meaning we'll need to update along to make sure ayon-blender works with older and newer releases.

We'll need to update all changed and deprecated API cases to support both Blender 4 and 5.

  

  

There also seems to be an issue with the Qt install on launch - that after launch QtPy still cannot launch accordingly. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1443216582116507670)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0246]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=bf69ff60cb8111f0b83443f824c7a70a)

Issue #201, opened 2025-11-26
Publisher - `bpy.context.scene.node_tree` removed in blender 5.0.0


Hisuuu:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Description
Since the release of Blender 5.0, the Python API changes the use of `Scene`, removing the access to `node_tree` (https://docs.blender.org/api/5.0/change_log.html#bpy-types-scene)

This deletion causes some bugs in the ayon Create and the ayon Publish, `ayon-blender` tries to use `bpy.context.scene.node_tree` sometimes.

Opening them will pop up a failed report:

<img width="854" height="371" alt="Image" src="https://github.com/user-attachments/assets/947391b2-b52f-4297-bf24-5880ce489065" />

And the `Collection Versions Loaded in Scene` publish plugin will also fail.

<img width="1299" height="828" alt="Image" src="https://github.com/user-attachments/assets/53a7f099-c829-4e86-b38f-035faeef1506" />

### Steps To Reproduce:

1. Launching Blender 5.0+ from the ayon launcher
2. Using the Create or the Publish as usual

### Version
* server : 1.13.0+202511071725
* ayon-core : 1.6.10
* ayon-blender : 1.0.10
* Blender : 5.0.0

### What platform were you running when you found the bug?
Linux - Rocky9

### Relevant log output:

```shell
Traceback (most recent call last):
  File "<path_to_ayon>/ayon/ayon-core/client/ayon_core/pipeline/create/context.py", line 1537, in reset_instances
    creator.collect_instances()
  File "/home/<user>/.local/share/AYON/addons/blender_1.0.10/ayon_blender/api/plugin.py", line 295, in collect_instances
    self.cache_instance_data(self.collection_shared_data)
  File "/home/<user>/.local/share/AYON/addons/blender_1.0.10/ayon_blender/api/plugin.py", line 215, in cache_instance_data
    if bpy.context.scene.node_tree:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Scene' object has no attribute 'node_tree'

---

Traceback (most recent call last):
  File "/home/<user>/.local/share/AYON/dependency_packages/ayon_2502101346_linux-rocky9.zip/dependencies/pyblish/plugin.py", line 528, in __explicit_process
    runner(*args)
[...]

PR #204, opened 2025-12-01
Support Blender 5.0+

BigRoy:
## Changelog Description

Add support for Blender 5.0

For reference, see: https://developer.blender.org/docs/release_notes/5.0/python_api

- Fix PySide install for Blender 5.0+
- Access comp node tree via `bpy.context.scene.compositing_node_group` now
- Refactor usd export argument that has been replaced in API.
- Refactor filepath logic for `CompositorNodeOutputFile` since API changes.

Fix: https://github.com/ynput/ayon-blender/issues/203
Fix: https://github.com/ynput/ayon-blender/issues/201

## Testing notes:
1. Integration should work with Blender 5.0
2. Integration should still work with Blender 4.5

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -779,33 +779,6 @@ def search_replace_render_paths(src: str, dest: str) -> bool:
     return changes
 
 
-def map_colorspace_name(colorspace: str) -> str:
-    """
-    Map ACES or other colorspace names to Blender's expected colorspace names.
-    DEPRECATED: This function is deprecated and will be removed in future
-    versions. Blender 5.0 natively supports ACES colorspaces.
-
-    Args:
-        colorspace: The original colorspace name
-
-    Returns:
-        str: The mapped colorspace name that Blender expects
-    """
-    colorspace_mapping = {
-        "ACES - ACEScg": "ACEScg",
-        "ACES - ACES2065-1": "ACES2065-1",
-        "ACES - sRGB": "sRGB",
-        "ACES - Rec.709": "Linear Rec.709",
-        "ACES - Rec.2020": "Linear Rec.2020",
-        "Linear": "Linear Rec.709",
-        "sRGB": "sRGB",
-        "Rec.709": "Rec.1886",
-        "Rec.2020": "Rec.2020",
-    }
-
-    return colorspace_mapping.get(colorspace, colorspace)
-
-
 def get_scene_node_tree(ensure_exists=False):
     """Return the node tree
 


client/ayon_blender/api/lib.py
```diff
@@ -771,3 +771,13 @@ def map_colorspace_name(colorspace: str) -> str:
     }
 
     return colorspace_mapping.get(colorspace, colorspace)
+
+
+def get_scene_node_tree():
```
iLLiCiTiT: Not sure what the overload of this is, but it might be good idea to cache it, with blender version too maybe (like if is above 5.0.0 boolean)?

BigRoy:
```python
Couldn't load user scripts "C:\Users\User\AppData\Roaming\Blender Foundation\Blender\5.0\scripts"
  File "F:\dev\ayon-blender\client\ayon_blender\api\lib.py", line 141, in append_user_scripts
    load_scripts(user_scripts.split(os.pathsep))
  File "F:\dev\ayon-blender\client\ayon_blender\api\lib.py", line 30, in load_scripts
    import bpy_types
ModuleNotFoundError: No module named 'bpy_types'
```
I do still get this at startup in Blender 5.0 - but I'm not sure what we want to do there. I'm a bit at loss why we are still in need to even manage Blender's users scripts paths, etc. Does anyone remember? @LiborBatek @moonyuet @antirotor perhaps?

TheLineAnimation:
> ```python
> Couldn't load user scripts "C:\Users\User\AppData\Roaming\Blender Foundation\Blender\5.0\scripts"
> Traceback (most recent call last):
>   File "F:\dev\ayon-blender\client\ayon_blender\api\lib.py", line 141, in append_user_scripts
>     load_scripts(user_scripts.split(os.pathsep))
>   File "F:\dev\ayon-blender\client\ayon_blender\api\lib.py", line 30, in load_scripts
>     import bpy_types
> ModuleNotFoundError: No module named 'bpy_types'
> ```
> 
> I do still get this at startup in Blender 5.0 - but I'm not sure what we want to do there. I'm a bit at loss why we are still in need to even manage Blender's users scripts paths, etc. Does anyone remember? @LiborBatek @moonyuet @antirotor perhaps?

I would say this is better handled by using the new [script_directories](https://docs.blender.org/manual/en/latest/editors/preferences/file_paths.html#script-directories) and sack off blender user scripts 

```
devFolder = bpy.context.preferences.filepaths.script_directories.new()
devFolder.name = "ROY"
devFolder.directory = [filepath]
```

BigRoy:
I disabled the loading of user scripts in Blender 5+ - The startup error is now gone.

@LiborBatek can you give this a test go please? @MilaKudr feel free to confirm the startup errors are gone and do more testing if needed.

LiborBatek:
@iLLiCiTiT  yes indeed still need refactoring , publishing works and the rest of tools also ... but `renders` are not functional due to massive changes in the `blender 5.0`

I will report more specificaly on rendering issues soon...sry been off for a while

LiborBatek (changes requested):
Overall `Blender 5.0` and Ayon integration works with all product types

all been tested and also Managing tools.

However rendering has issues due to changes in this blender version.

Here is compositor with tree in `blender 4.5.3` 



Screenshot 2025-12-11 175440

Note correctly connected `Composite` node out into `File Output Node` used for rendering.

while in `blender 5.0` it completely lacks `Compositor` node and using `Group Out node` instead so our Ayon integration fail connecting such input into `File Output Node` marked by red:

Screenshot 2025-12-11 175820

However it seems that its working also with multiple `ViewLayers` aka render layers but just not with compositor out which is still unconnected.


Screenshot 2025-12-11 180127

And lastly I had issues with rendered frames which are empty and has no any `AOVs`...but that could also be my legacy scene configuration and need to be revisited more...


Screenshot 2025-12-11 150138

It also seems that even the `File Output Node` of new version changed too...

LiborBatek:
> @LiborBatek I've disabled the "Composite" option in Blender 5.0 because it does not have a "Composite" node.
> 
> Was there still anything else?

I will give it another testing round and see if Im able to produce "working" aka correct rendered images...and let you know!

LiborBatek:
if my scene having the `Input Scene` then it partially works and `Render` instance got created but the output node is not anyhow connected and without any reasonable inputs...again failure.

image

BigRoy:
```
IMB_exr_begin_write: ERROR: Cannot open image file "C:\projects\Blender\Shots\sh040\work\lighting\renders\blender\Blnd_sh040_lighting_v0031005.exr". Missing or empty channel list in header
```
That in your render logs doesn't seem too optimal. Unrelated?

---

> And couple of scrnshots of my out dir (staging work render directory with some renders in wrong place)

@LiborBatek can you share the logs of the Publish job in Deadline? Did it try to run once? If so, can you get me the log of the first run? I have a vague feeling it did end up publishing, but errored due to something else... so the files in the v003 render folder got removed but job still failed due to something else?

If that's not the case and it is actually rendering to the wrong folder - can you pass me that workfile? This I seem to have a hard time reproducing.

The rest preparing render setup I've now fixed.

PR #211, opened 2025-12-06
Fixes colorspace override in render collection

### LINKED ISSUE #210: Collect Render fails on file output colorspace override

## Current Behavior
Turning on override on Compositor file output leads to Collect Render Exception

### Steps To Reproduce:

Image

### Version
Version of the server 1.13.0, Core 1.6.7 Blender 1.0.9

Windows

### Relevant log output:

```shell

File "C:\_GFX_library\ayon\dependency_packages\ayon_2510271030_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process

runner(*args)

File "C:\_GFX_library\ayon\addons\blender_1.0.9\ayon_blender\plugins\publish\collect_render.py", line 131, in process

colorspace_data = self.get_colorspace_data(comp_output_node)

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

File "C:\_GFX_library\ayon\addons\blender_1.0.9\ayon_blender\plugins\publish\collect_render.py", line 152, in get_colorspace_data

display: str = node.display_settings.display_device

^^^^^^^^^^^^^^^^^^^^^

AttributeError: 'CompositorNodeOutputFile' object has no attribute 'display_settings'
```

jrsndl:
## Changelog Description
Ensures the colorspace setting is correctly overridden when color management is enabled in the render node, fetching the colorspace from the linear colorspace settings.

Fix #210 

Blender 4.5

### DIFF
--- a/client/ayon_blender/api/capture.py
+++ b/client/ayon_blender/api/capture.py
@@ -121,8 +121,12 @@ def isolate_objects(window, objects):
     """Isolate selection"""
     deselect_all()
 
+    # Only select objects that happen to be in the view layer to avoid
+    # errors that object can't be selected because it is not in View Layer.
+    view_layer_objects = window.view_layer.objects
     for obj in objects:
-        obj.select_set(True)
+        if obj.name in view_layer_objects:
+            obj.select_set(True)
 
     context = create_blender_context(selected=objects, window=window)
 


jrsndl:
This PR also adds the colorspace of the file. I assume the intent of the override is to actually collect that, and not only display  & view.
I set the default to be ACEScg, but maybe it would be better to have a workfile settings for Blender in a similar fashion to Maya, for setting up the Rendering Space:

image

BigRoy (approved):
Thanks @jrsndl this works.

I removed the magical `ACEScg` fallback and added todos in the code where I felt it's still lacking or potentially incorrect in a certain case.

I've also removed the fake fallback data if not color managed - since that's allowed with core >=1.6.10 so update the requirements accordingly.

Would be great if you can test it some more @jrsndl 
Other than that, @LiborBatek can you give this a whirl?

Issue #213, opened 2025-12-08
YN-0295: Blender: Create camera instance reparents camera affecting its transformation


ynbot:
### Please describe the issue you have encountered?

If you have a camera and the parents have transformations and you select camera, create camera instance... well then it reparents it to an "asset group" and goodbye goes your good old parent that had some nice animation for the camera.

Essentially we want to preserve the exact hierarchy that the user had for the camera, not move the camera to another parent.

### How to replicate the issue?

1. Create camera
2. Add some nice parents
3. Move the parents, key them nicely.
4. Select camera
5. Create camera instance.

And gone is all the motion it got from its previous parents.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0295]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=7befc4d0d43d11f0aabfb51851f62e18)

PR #214, opened 2025-12-08
Camera: Use Collection instead of Asset Group for Camera instance

BigRoy:
## Changelog Description

Use Collection instead of Asset Group for creating Camera instances, this way the parent hierarchy remains intact and it disallows the camera to lose e.g. parent transformations, etc.

Fix: https://github.com/ynput/ayon-blender/issues/213

## Testing notes:
1. Legacy camera instances should still work
2. New camera instances should work
    - please also check whether the output hierarchy after export is the way you'd expect.
    - please also check whether this may affect other stuff you can think of @LiborBatek @moonyuet

### DIFF
--- a/client/ayon_blender/plugins/load/load_camera_fbx.py
+++ b/client/ayon_blender/plugins/load/load_camera_fbx.py
@@ -68,9 +68,10 @@ class FbxCameraLoader(plugin.BlenderLoader):
             ayon_info = obj[AYON_PROPERTY]
             ayon_info.update({"container_name": group_name})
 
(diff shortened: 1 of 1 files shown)


BigRoy:
> > @LiborBatek I can't seem to reproduce this - FBX load works fine for me:
> > image
> > Tested with Blender 4.5.
> 
> I have used parent `locator` for animation so the `camera` is nested inside of such relationship/hierarchy... yours having just a camera afaik

My source scene:

image

On load with FBX:
image

Seems absolutely fine?

Issue #208, opened 2025-12-05
YN-0280: Blender: Add support for local render


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Add support for local rendering without requiring to submit to farm 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1443201943127330911)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0280]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=378a3af0d03711f0bf376d0529e683b0)

PR #217, opened 2025-12-10
Render: Add support for local rendering

BigRoy:
## Changelog Description

Render: Add support for local rendering

Fix https://github.com/ynput/ayon-blender/issues/208

## Testing notes:

1. Farm render and local render should work

### DIFF
--- a/client/ayon_blender/plugins/publish/collect_instance.py
+++ b/client/ayon_blender/plugins/publish/collect_instance.py
@@ -91,11 +91,11 @@ class CollectBlenderInstanceData(plugin.BlenderInstancePlugin,
         return [
             BoolDef(
                 "collection_include_object_children_recursive",
-                label="Include Objects Hierarchy",
+                label="Expand to All Children",
                 default=True,
                 tooltip=(
-                    "If enabled, the children of objects in the collection "
-                    "will be included in the instance members.\n"
+                    "Includes all descendant children of the linked objects, "
+                    "even if not manually selected.\n"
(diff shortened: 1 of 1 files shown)


LiborBatek:
@BigRoy yep, sure... here you have it.

[B3d-LocalRender-publish-report-251210-15-00.json.txt](https://github.com/user-attachments/files/24160864/B3d-LocalRender-publish-report-251210-15-00.json.txt)

BigRoy:
> @BigRoy I was able to render locally and the images successfully got integrated... there was a strange behaviour tho when rendering on farm as it was collecting wrong version for the product (v004 instead of v006), Im really not sure if its connected to this PR or rather issue in the `ayon-blender` in fact. I need to investigate it more...but maybe I faced a specific scenario as Im having both multilayered exrs products and singlelayered exrs too (as I have performed multiple tests already)...

Hard to tell - I'll just wait until you're confident to approve 👼

client/ayon_blender/api/plugin.py
```diff
@@ -446,7 +450,12 @@ def read(self, node) -> dict:
         if not ayon_property:
             return {}
 
-        return ayon_property.to_dict()
+        data = ayon_property.to_dict()
+        data["families"] = self.get_publish_families()
+        return data
+
+    def get_publish_families(self) -> list[str]:
```
iLLiCiTiT: Can we add new families in collection phase instead? What is the point of storing them to instance metadata when they can be added during publishing...

PR #219, opened 2025-12-11
Alembic Camera Loader: Object path would be always filled during versioning

moonyuet:
## Changelog Description
This PR is to make sure object path is not empty during versioning in scene inventory.

n/a

## Testing notes:
1. Load Camera (Alembic)
2. Update to the latest or set version in scene inventory
3. Object path should be filled.


LiborBatek (approved):
Now all been working well and I was able to switch `camera` and Abc repre without any issues,

LGTM

PR #220, opened 2025-12-11
Fbx Camera Loader: No duplicated objects linked to scene collections

moonyuet:
## Changelog Description
This PR is to provide a quick fix to make sure no duplicated objects linked to scene collections. 

As there is no duplicated object linked to scene collections.

## Testing notes:
1. Load Camera (both abc and fbx)
2. Set version or update to the latest
3. The error should be gone

### DIFF
--- a/client/ayon_blender/plugins/load/load_camera_fbx.py
+++ b/client/ayon_blender/plugins/load/load_camera_fbx.py
@@ -53,6 +53,13 @@ class FbxCameraLoader(plugin.BlenderLoader):
             obj.parent = asset_group
 
         for obj in objects:
+            if obj.name in parent.objects:
+                self.log.info(
+                    "Object with name '%s' already exists in the scene. "
+                    "Skipping linking to the scene collection.",
+                    obj.name,
+                )
+                continue
             parent.objects.link(obj)
             collection.objects.unlink(obj)
 


client/ayon_blender/plugins/load/load_camera_fbx.py
```diff
@@ -53,6 +53,13 @@ def _process(self, libpath, asset_group, group_name):
             obj.parent = asset_group
 
         for obj in objects:
+            if obj.name in parent.objects:
+                self.log.info(
+                    "Object with name '%s' already exists in the scene. "
+                    "Skipping linking to the scene collection.",
+                    obj.name,
+                )
+                continue
```
BigRoy: Could you elaborate how this fixes it? It seems like a hacky 'fix'?

BigRoy: > (as it happens when users load a camera in the same instance I guess).

Can you explain how I would reproduce this? How do I load a camera to the same instance?

BigRoy: But aren't we namespacing everything? 🤔 I'm so confused.

@LiborBatek can you share a screen recording of it breaking without this PR?

LiborBatek:
@BigRoy here you have it while in `ayon-blender`  `1.0.11` ...those products originates in `maya` but I guess it doesnt make any difference to any blender based ones too..

https://github.com/user-attachments/assets/a3c6d61d-ebb8-4741-8671-aafa4460e5a2

moonyuet:
@LiborBatek can you test with the latest develop and see if the issue has been resolved?

Issue #224, opened 2025-12-16
bpy.app.timers _assign_image_to_node did not return None or float


alainxi:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Description

This appears in the Blender console :
Error: 'bpy.app.timers' callback <function LoadImageShaderEditor._assign_image_to_node.<locals>.<lambda> at 0x000001823FB80680> did not return None or float.

It happens each time I use one of these : 
- Ayon-->Load-->Load to Shader Editor
- Ayon-->Manage-->Switch Folder-->Load to Shader Editor

It started since the release of this PR in ayon-blender 1.0.11 : https://github.com/ynput/ayon-blender/pull/187

I guess the problem is near this line :  https://github.com/ynput/ayon-blender/blob/5b3832c5d0cfc2c43202c5cf9d753064c3e5cc94/client/ayon_blender/plugins/load/load_image_shader.py#L206

### Steps To Reproduce:

Described in the text above.

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #227, opened 2025-12-17
Load image: bpy.app.timers return float to retry the timer 

moonyuet:
## Changelog Description
This PR is to fix the bug encountered when loading images with shader editor and switching it to other texture. 
```
Error: 'bpy.app.timers' callback <function LoadImageShaderEditor._assign_image_to_node.. at 0x000001823FB80680> did not return None or float.
```
Resolve https://github.com/ynput/ayon-blender/issues/224

n/a

## Testing notes:
1. Ayon-->Load-->Load to Shader Editor
2. Ayon-->Manage-->Switch Folder-->Load to Shader Editor

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -665,7 +665,7 @@ def strip_namespace(containers):
 
 
 @contextlib.contextmanager
-def packed_images(datablocks):
+def packed_images(datablocks, logger=None):
     """Unpack packed images during context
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
I have tested with latest `ayon-blender` which trigger that error and this branch and it stopped showing that error in the console...LGTM

Before:
image


After:
Screenshot 2025-12-18 110037

I have also tested when version managing the file texture and also no errors shown while doing so....

PR #229, opened 2026-01-09
Expand to All Children: Improve label and tooltip

BigRoy:
## Changelog Description

Change label `Include Objects Hierarchy` to `Expand to All Children` since it feels less confusing to the users + tweak the wording on the tooltip.

Based on user feedback asking questions about what it does.

## Testing notes:

1. Check whether it's better.


LiborBatek (approved):
Works for me!
image

Issue #223, opened 2025-12-13
YN-0227: Blender - Incorporate implicit Animation product when loading a Rig


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Based on clients Catchup call on 18/11/2025 with Marian, Jiri and Baji (who brought this idea to the table).

Currently we are missing auto creation of

```
Animation
```

product inside Blender host. We need similar approach as in e.g. Maya where this product been automatically created and named when loading Rig.

  

Namely due to possible confussion and/or inconsistencies while created manually by artists themselves...as there could be variants differences within shot or among multiple shots which is not desirable... 

Based on clients Catchup call on 18/11/2025 with Marian, Jiri and Baji (who brought this idea to the table).

Currently we are missing auto creation of

```
Animation
```

product inside Blender host. We need similar approach as in e.g. Maya where this product been automatically created and named when loading Rig.

  

Namely due to possible confussion and/or inconsistencies while created manually by artists themselves...as there could be variants differences within shot or among multiple shots which is not desirable...

### Additional context:

_This issue was automatically created from AYON ticket [YN-0227]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=0d83a4a0c51d11f0810f15cd4c163fb2)

PR #233, opened 2026-01-13
Add options to auto create animation instance for rig product

moonyuet:
## Changelog Description
This PR is to support the options of auto-creating animation instance for loaded rig product in blender. The option can be turned on/off through the ayon settings.
Fixes: https://github.com/ynput/ayon-blender/issues/223

Discussion needed for whether it should be also incorporated into the blendScene product

## Testing notes:
1. Turn on/off `ayon+settings://blender/load/BlendLoader/create_animation_instance_on_load`
2. Load your rig product by Append Blend in blender
3. If you turn on the auto-creation, it would automatically create animation instance for you if your product type is rig or there is a rig in your layout product.


client/ayon_blender/api/lib.py
```diff
@@ -803,3 +803,24 @@ def get_scene_node_tree(ensure_exists=False):
...
+    creator_identifier = "io.ayon.creators.blender.animation"
+    host = registered_host()
+    create_context = CreateContext(host)
+
+    create_context.create(
+        creator_identifier=creator_identifier,
+        variant=rig.name.split(':')[-1],
+        pre_create_data={
+            "use_selection": False,
+            "asset_group": rig
```
BigRoy: What does passing the `asset_group` do? :)

LiborBatek (approved):
I have tested in `blender 4.5.3` and all been working nicely!

I was able to publish succesfully, load such `animation` product into new scene and also version manage such product.

There is one small thing to adjust but more of a cosmetics and thats the instance name which need some underscore separator in the name...as now its all together as seen on a pic below:
image


Disabling of `auto creation of animation` works well too...

LGTM

Issue #234, opened 2026-01-14
Incorporation of Templated Workfile Builder


moonyuet:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
We have the settings for workfile builder but it does not fully integrate with workfile builder from the Core Addon. 

## Expected Behavior
We need to fully integrate the workfile builder system with menu and inherited logic from the workfile builder

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #239, opened 2026-01-19
Implementation of the workfile template builder

moonyuet:
## Changelog Description
This PR is to add the missing workflow of work template build in blender and to fully incorporate the implementation. 
Resolves https://github.com/ynput/ayon-blender/issues/234

n/a

## Testing notes:
1. Launch Blender
2. Tweak your setting in `ayon+settings://blender/templated_workfile_build/profiles/0`
3. `Menu->Templated Workfile -> Build Workfile from Template`

### DIFF
--- a/client/ayon_blender/api/plugin_load.py
+++ b/client/ayon_blender/api/plugin_load.py
@@ -160,8 +160,8 @@ def link_collection(
         bpy.types.Collection: The loaded collection datablock.
     """
     asset_container = get_collection(group_name)
-    # Rule: remove the second token if it is exactly two digits (e.g. "_01_").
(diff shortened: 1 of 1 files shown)


moonyuet:
As I and @LiborBatek discussed on the call, we would implement the current templated workfile builder workflow in this way. 
1. We would implement the first workfile creation first as we heard some users need this for loading template and save as first workfile(This is the initial idea about this PR)
2.  We would comment out update workfile from template first as updating those meaning that we need to also implement the load placeholder
3.  We would include the options of `Open Template` 
4. We would place the workfile menu to the top to differentiate from the `Templated Workfile`

We would create issues about implementation of load placeholders and updated template from workfile, so once it becomes sponsored or we have time, we can discuss on how to implement those.

PR #247, opened 2026-01-29
Link Blend Flat won't link to the already linked collection

moonyuet:
## Changelog Description
This PR is to make sure Link Blend Flat won't link to the already linked collection again and error out as `bpy.op.wm.link` is already set the active_collection to True

n/a

## Testing notes:
1. Load asset from Link Blend Flat 
2. Should be working


LiborBatek (approved):
I have tested with both Link loaders and all good!

LGTM

PR #248, opened 2026-01-30
Link Blend Flat: Improve Regex to find the collection name

moonyuet:
## Changelog Description
- Support to load collection where the project has folder name which starts/ends with number

Tested by client in production 

## Testing notes:
1. Create task within the folder which named either `00_shot01` or `dev_01`
2. Launch Blender in these tasks
3. Create BlendScene
4. Publish
5. Link Blend Flat


BigRoy (commented):
I still think it's worth investigating whether we can get by without the whole name matching - it'll be fragile now, and always will be. We just can't assume something will have and remain having the same name over time.

Are we sure there isn't a way we can actually load this data and manage it directly using that piece of data - instead of having to rely on expecting some names somewhere?

Issue #249, opened 2026-02-02
Blender 5.1 support


nebukadhezer:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
[This will break in 5.1 due to py3.13.](https://github.com/ynput/ayon-blender/blob/develop/client/ayon_blender/hooks/pre_pyside_install.py#L85)

## Expected Behavior
Install the latest PySide6 bindings

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

krishnaavril:
Tested with Rocky 9 with current new blender and core addon, its not working!

BigRoy:
> its not working!

Not working where - please provide more details ;)

krishnaavril:
ahh sorry, I could have given more info

we are not unable to see the ayon menu in the tab with 5.1, below is the console log

```
Traceback (most recent call last):
  File "/mnt/netapps/Blender/blender-5.1.0-linux-x64/5.1/scripts/modules/bpy/utils/__init__.py", line 142, in _test_import
    mod = __import__(module_name)
  File "/home/avril9950x3d/.local/share/AYON/addons/blender_1.1.1/ayon_blender/blender_addon/startup/init.py", line 3, in <module>
    from ayon_blender.api import BlenderHost
  File "/home/avril9950x3d/.local/share/AYON/addons/blender_1.1.1/ayon_blend00:06.255  color_management | WARNING scene view "AgX" not found, setting to "sRGB".
00:06.255  color_management | WARNING scene output view "Standard" not found, setting to "sRGB".
00:06.255  color_management | WARNING sequencer colorspace "sRGB" not found, will use default instead.
00:17.808  blend            | Saved session recovery to "/tmp/quit.blend"
register_class(...):
Info: Registering key-config preferences class: 'Prefs', bl_idname 'Blender' has been registered before, unregistering previous
register_class(...):
Info: Registering key-config preferences class: 'Prefs', bl_idname 'Blender' has been registered before, unregistering previous

Blender quit
```

BigRoy:
Can you launch ayon-blender via CLI and see what the pre launch hook tells you?

E.g. with older `ayon-blender` I got:
```
[...]

PR #250, opened 2026-02-02
Chore: Update PySide version

nebukadhezer:
…issues with a specific version of PySide6

## Changelog Description
Remove the PySide6 version pin to 6.6.3. 
Newer PySide6 versions like 6.10.x run the code below in blender.
Blender 5.1 fails with the version pin to 6.6.3 as there is no version on pip for py3.13 which is 5.1 default I guess.

https://github.com/ynput/ayon-core/pull/398

this will close 
https://github.com/ynput/ayon-blender/issues/249

## Testing notes:
```
from PySide6 import QtWidgets
layout = QtWidgets.QHBoxLayout()
widget = QtWidgets.QWidget()
text_edit = QtWidgets.QTextEdit()
# This works
layout.addwidget(widget)
# This crashes because of "wrong type"
layout.addwidget(text_edit)
new = QtWidgets.QWidget()
new.setLayout(layout)
new.show()
```

### DIFF
--- a/client/ayon_blender/api/capture.py
+++ b/client/ayon_blender/api/capture.py
@@ -163,9 +163,8 @@ def _apply_options(entity, options):
 def applied_view(window, camera, isolate=None, options=None):
     """Apply view options to window."""
     area = window.screen.areas[0]
-    space = area.spaces[0]
-
     area.ui_type = "VIEW_3D"
+    space = area.spaces[0]
 
     types = {"MESH", "GPENCIL"}
     objects = [obj for obj in window.scene.objects if obj.type in types]
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -350,6 +350,108 @@ def get_selection(include_collections: bool = False) -> List[bpy.types.Object]:
     return selection
 
 
+def get_upstream_viewlayers(node: bpy.types.CompositorNodeOutputFile)-> set[str]:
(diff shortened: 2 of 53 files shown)


BigRoy:
I tested in Blender 4.1.1, Blender 4.5 and Blender 5.0 - each of those currently downloaded PySide6 version `6.10.2` with this PR. 

They all worked, as in - they did not crash. However, it did stand out to me that this list of Creators suddenly came with massive top and bottom padding?

_Also note list widget being much further indented.

image

And the publish report started looking all funky:
image

Both these issues do not occur with PySide 6.6.3:

image

image

As such, it won't crash if we update but we have some major styling issues then that I think we should tackle before merging this.

(Tagging @philippe-ynput @iLLiCiTiT for visibility)

@nebukadhezer would you like the puzzle of trying to figure out what changes are needed in the core stylesheet or `ayon-core` code to make it work with that newer PySide in Blender? If so - it'd be greatly appreciated.

Also, if we can get it to work, I'd pin the PySide version just to the newer one to avoid anyone in community or anyone in a particular studio running Blender for the first time at a later point potentially getting different UI issues, etc.

iLLiCiTiT:
I would lock the version to be honest. Or add settings for having option to not lock it.

We did lock it for a reason in past, latest release had a bug, which caused "it suddenly stopped working", which is the worst thing to happen and debug.

iLLiCiTiT (approved):
Looks good to me

Issue #235, opened 2026-01-15
YN-0397: Blender: Introduce a `Look` product type


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

  

Currently there is no dedicated product in Blender for

```
Look
```

so we should use e.g. blendScene type and tweak it so it produces a publish consisting of just material and linked datablocks like image textures only. Then be able to load and version manage such a product. This feature and possible resulting PR should not include any assignment to the scene entities but only incorportion of this product alone. There gonna be other ticket for the full lookdev workflow... 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0397]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=85ebcde0f20f11f0b6aac71a3095140f)

BigRoy:
I've assigned this to you @moonyuet 

After discussing today with Philippe and Libor this part is the 'barebones' of exporting the materials and loading back in the databalocks.

The idea is: 
- Creator creates a collection that would contain the objects for which to publish the look (similar to maya). So the user just selects from which objects to export their looks.
- On publish, for each of the objects we collect the materials and write out a `.blend` file with only these material datablocks.
- A loader gets added that just links this `.blend` and loads all `material` datablocks.
- The loader containerises the library - so what the loader manages, is just the library path (not sure how feasible that is?)
- Preferably on update any new materials also become available in the same link.

---

Even though not entirely what this ticket is about I did want to also share this [cool look workflow video for Blender](https://www.youtube.com/watch?v=IBTEBhAouKc&t=509s) that Libor shared. It's the video we mostly discussed in the call earlier today - and from which we decided for now (due to this being the smaller starter ticket) to just do a first pass on only managing the material datablocks.

PR #262, opened 2026-02-20
Implementation of Look Product Type

moonyuet:
## Changelog Description
This PR is to implement the look workflow in blender which is similar to that in maya.
Resolve https://github.com/ynput/ayon-blender/issues/235

We should consider simplify and make parent class for all those datablock linkers (action and material)

## Testing notes:
1. Create Look 
2. Publish
3. The related material datablock should be Loaded with `Load Look`

### DIFF
--- a/client/ayon_blender/api/lib.py
+++ b/client/ayon_blender/api/lib.py
@@ -376,6 +376,12 @@ def iter_images_in_node_tree(tree: bpy.types.NodeTree):
 
 @contextlib.contextmanager
 def make_material_image_paths_absolute(material_datablocks: set[bpy.types.Material]):
+    """Make image paths in materials absolute during context.
+
+    Args:
+        material_datablocks (set[bpy.types.Material]): material datablocks to make 
+            image paths absolute for during context
+    """
     original_image_paths = {}
     for material in material_datablocks:
         if not material.use_nodes:


client/ayon_blender/plugins/create/create_look.py
```diff
@@ -0,0 +1,37 @@
...
+            product_name, instance_data, pre_create_data
+        )
+        if pre_create_data.get("use_selection"):
+            for obj in lib.get_selection():
+                if not obj.active_material:
+                    empty_material = bpy.data.materials.new(name=product_name)
+                    empty_material.use_nodes = True
+                    obj.data.materials.append(empty_material)
+                else:
+                    obj.active_material.name = product_name
```
BigRoy: Please do not touch the user's scene - we just want to mark the objects from which to collect materials, we don't want to change the materials.

I was so confused why my materials were suddenly renamed.


```suggestion
```

client/ayon_blender/plugins/load/load_look.py
```diff
@@ -117,109 +60,85 @@ def process_asset(
...
-        selected = [o for o in bpy.context.scene.objects if o.select_get()]
-
-        materials, objects = self._process(libpath, container_name, selected)
-
-        # Save the list of imported materials in the metadata container
-        metadata["objects"] = objects
-        metadata["materials"] = materials
+        container_metadata = container.get(AYON_PROPERTY)
+
+
```
BigRoy: Check your linter please.
```suggestion
```

client/ayon_blender/plugins/load/load_look.py
```diff
@@ -117,109 +60,85 @@ def process_asset(
...
-            f"The file doesn't exist: {libpath}"
-        )
-        assert extension in VALID_EXTENSIONS, (
-            f"Unsupported file: {libpath}"
-        )
+        This will remove all objects of the current collection, load the new
+        ones and add them to the collection.
+        If the objects of the collection are used in another collection they
+        will not be removed, only unlinked. Normally this should not be the
+        case though.
```
BigRoy: Docstring sounds wrong - that's not what it should be doing. It should only be swapping a filepath on the datablock, right?

client/ayon_blender/plugins/create/create_look.py
BigRoy: > But I am pretty sure you need to have the material assignment if there is no material found in your mesh.

For what? We have a validator to tell the user if it does not have a material. Right? :) It's up to the user to create their looks, we're just extracting them from the scene files.

client/ayon_blender/plugins/load/load_look.py
```diff
@@ -117,109 +60,85 @@ def process_asset(
...
+        empty_obj.active_material = container
+        empty_obj.active_material.use_fake_user = options.get(
+            "use_fake_user", True
+        )
+        # Save the list of objects in the metadata container
+        container_metadata["libpath"] = libpath
+        container_metadata["lib_container"] = lib_container
+        container_metadata["objects"] = empty_obj
+        container_metadata["material"] = empty_obj.active_material
+        empty_obj.data.materials.append(empty_obj.active_material)
```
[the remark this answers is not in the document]
moonyuet: Do we need to think about whether the look assignment should be per material datablock? We did the similar thing with action product type which containerises per action datablock, Isn't that would be easier to be managed and perhaps possible refactoring to make the two product types based on one loader?

client/ayon_blender/plugins/load/load_look.py
```diff
@@ -117,109 +60,72 @@ def process_asset(
...
+            context (Dict): The context of the update.
+        """
+        repre_entity = context["representation"]
+        collection = container["node"]
+        libpath = self.filepath_from_context(context)
+        materials = container["materials"]
+        for material in materials:
+            if material.library:
+                material.library.name = os.path.basename(libpath)
+                material.library.filepath = libpath
```
BigRoy: Do all the materials share the same `library`? If so, can't we just imprint the `library` itself, or why not? Like, can the library itself be the container?

client/ayon_blender/plugins/publish/validate_no_material.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @classmethod
+    def get_invalid(cls, instance):
+        invalid = []
+        for obj in instance:
+            if not (
+                isinstance(obj, bpy.types.Object)
+                and obj.type in {
+                    "MESH", "MATERIAL"
+                }
```
BigRoy: This is wrong - it should also allow volumes, etc.
```suggestion
```

client/ayon_blender/plugins/publish/extract_blend.py
```diff
@@ -145,3 +145,31 @@ def add_datablock(self, instance: pyblish.api.Instance) -> set:
...
+        """
+        materials = set()
+        for obj in instance:
+            if not isinstance(obj, bpy.types.Object):
+                continue
+            if not hasattr(obj.data, "materials"):
+                continue
+            if obj.active_material:
+                materials.add(obj.active_material)
+        return materials
```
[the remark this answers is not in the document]
moonyuet: Isn't the active material the one assigned to the mesh?

client/ayon_blender/plugins/load/load_look.py
```diff
@@ -73,7 +72,12 @@ def process_asset(
                 "there is any material datablock in the blend file."
             )
         materials = data_to.materials
-        container_metadata["materials"] = materials
+        container_metadata["libraries"] = [
+            material.library for material
+            in materials if material.library
+        ]
```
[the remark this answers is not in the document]
BigRoy: Explain. What's the doubts? What would fail?

LiborBatek (approved):
I was able to create, publish, load and version manage such `Look` product successfully.

As initial integration it looks good to me!

Here is my `Look` product loaded and visible in the `Scene Inventory`

Image

LiborBatek:
@BigRoy even your findings are correct Im still wondering if someone would use linked materials for lookdev or rather while in the lookdev workfile before publishing a `Look`, as I think it does not make much sense as its the source of such material or look product...so user either create it or append it from elsewhere...so its questionable whether to allow such circumstances or rather validate it as forbidden. I would rather go for simpler one (not allow such occasion?)

client/ayon_blender/plugins/publish/extract_blend.py
```diff
@@ -145,3 +145,40 @@ def add_datablock(self, instance: pyblish.api.Instance) -> set:
...
+        datablock_to_be_exported = set()
+        for obj in instance:
+            if not isinstance(obj, bpy.types.Object):
+                continue
+            if not hasattr(obj.data, "materials"):
+                continue
+
+            for material in obj.data.materials:
+                if material.library:
+                    material.make_local(clear_asset_data=True)
```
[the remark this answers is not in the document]
BigRoy: See: https://github.com/ynput/ayon-blender/pull/262#issuecomment-3986638840

Please do not enforce the local bit on export, but perhaps make a validator instead.

LiborBatek (changes requested):
Now the `Look` product doesnt contain any packed textures and while loading such `Look` product into new scene it simply been missing textures as they are mapped to the `.../publish/textures/` but there are no any files located after publishing the look in preceding step before loading the look product...as seen below

Image

while in previous commits before changes the texture resources been packed / self contained inside the `Look` product so no missing:

Image


We should copy the source textures to the publish look folder as the filepath suggests atm and introduce optional `Pack Textures` function on `Look` publish instance too 

or simply pack those always (for now) which is questionable...

I will leave it to you @BigRoy  @moonyuet

client/ayon_blender/plugins/publish/extract_blend.py
```diff
@@ -145,3 +145,37 @@ def add_datablock(self, instance: pyblish.api.Instance) -> set:
...
+        datablock_to_be_exported = set()
+        for obj in instance:
+            if not isinstance(obj, bpy.types.Object):
+                continue
+            if not hasattr(obj.data, "materials"):
+                continue
+
+            # We want to export all materials linked to the objects in the instance,
+            # to preserve the node tree and texture links.
+            datablock_to_be_exported.add(obj)
```
BigRoy: I see now @LiborBatek mentioning packed textures fail - but if we can we should really avoid writing the object. I'm actually quite surprised that packing textures requires the object.

I think the issue we're facing is that the used `image` datablocks used by the materials are not being written out. That should contain the packed images, not the object.

Please figure out an approach where we can avoid the need for the object datablock. I don't see any reason why Blender would need the object to have the image/material datablocks - even if packed 🤔

BigRoy:
> We should copy the source textures to the publish look folder as the filepath suggests atm and introduce optional Pack Textures function on Look publish instance too
> 
> or simply pack those always (for now) which is questionable...

Copying over textures - may be nice. I think best case would be a Textures enum or setting that would decide between:
- Pack (pack into blend file)
- Use original paths (do not alter the datablocks)
- Copy with publish (similar to maya look resources)

moonyuet:
@BigRoy @LiborBatek 
It seems that I have misunderstood a little bit. I have found the root cause of the issue and this could be the issue where we have similar testing scene but with different result.
When we set blender for `bpy.ops.file.make_paths_absolute`, and run the publish, it will always work to publish material datablock correctly with the absolute path of the texture. As it would make all the paths absolute.
But users might set blender for `bpy.ops.file.make_paths_relative` and runt the publish, it will always publish the material with relative path, which means even if we added the resource path, we can't manage to link back the right material as we set the texture filepath to relative.
The fix I committed just now is to make sure all the texture paths are absolute paths during context of publishing datablock

client/ayon_blender/api/lib.py
```diff
@@ -349,6 +349,46 @@ def get_selection(include_collections: bool = False) -> List[bpy.types.Object]:
...
+            processed.update(children)
+
+@contextlib.contextmanager
+def make_path_absolute(material_datablocks: set[bpy.types.Material]):
+    original_image_paths = {}
+    for material in material_datablocks:
+        if not material.use_nodes:
+            continue
+        for node in iter_images_in_node_tree(material.node_tree):
+            image = node.image
```
BigRoy:
```suggestion
        for image in iter_images_in_node_tree(material.node_tree):
```

PR #263, opened 2026-02-23
Backward compatibility for blender pyside install settings

moonyuet:
## Changelog Description
This PR is to add backward compatibility for blender pyside install settings.
 
Draft as it's only necessary for the dev mode. 
But no guarantee there is some edge case.

## Testing notes:
1. Launch Blender
2. Run successfully


iLLiCiTiT:
This is something that can happen only to developers. I would really avoid doing this, I know it is PITA, but many times happened to us that we did use these `.get(` and then found out it never uses the settings because one of the keys is different, or the settings were removed, it really hides a lot of potential issues. Access to settings should be always explicit.

PR #264, opened 2026-02-24
Fix BlenderLoader._load return values.

rdelillo:
## Changelog Description

Fix `BlenderLoader._load` return value.

## Testing notes:
1. Should still work as previously

### DIFF
--- a/client/ayon_blender/api/plugin.py
+++ b/client/ayon_blender/api/plugin.py
@@ -562,16 +562,12 @@ class BlenderLoader(LoaderPlugin):
             folder_name, product_name, unique_number
         )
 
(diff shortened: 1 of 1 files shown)


client/ayon_blender/api/plugin.py
```diff
@@ -562,17 +562,13 @@ def _load(self,
...
+        return self.process_asset(
             context=context,
             name=name,
             namespace=namespace,
             options=options,
         )
 
-        # Only containerise if anything was loaded by the Loader.
-        if not nodes:
-            return None
```
rdelillo: code was returning `None` anyway.

Issue #265, opened 2026-02-24
YN-0521: Blender: Load Cache allow disabling namespaces on load


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

When loading caches it will automatically prepend loaded data with a namespace like '{group_name}:' - there may be cases where you don't want this. So please make this optional. 

### Additional context:
link to discussion on [Discord](https://sonic4-vfx.slack.com/archives/C0A9SBY6WSW/p1771859991971499)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0521]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=8104560011d411f1b09e01f2f5f577df)

PR #266, opened 2026-02-24
Load Cache: Allow disabling namespace on load

BigRoy:
## Changelog Description

Allow disabling group namespace on load for loading Alembic caches/cameras

Allow to load without namespaces applied.

Fix: https://github.com/ynput/ayon-blender/issues/265

## Testing notes:

1. Settings and load should respect the load behavior to disable add namespaces, etc.


moonyuet (approved):
Looks good to me but should we just change `group_name` to `namespace` instead? The original `namespace` variable in these two loaders can only be found in the container data and it is not used anywhere.

PR #283, opened 2026-04-07
Load Cache: Do not remove materials if there are object linking to that during update

### LINKED ISSUE #282: YN-0641: Blender: Load Cache removes materials

In Blender using the Load Cache loader it seems that when you're updating or removing that it also removes the assigned materials from the objects, which is uwanted behavior _especially_ on update.

<br>

Possibly the best approach is to not touch the materials at all, or find a way that we can ensure we only touch materials that were loaded by the loader itself (which I'm not sure whether we can reliably?)

<br>

Deleting just materials that have no "users" may also not be entirely accurate because after e.g. "update" it may be that the material seems unused but should've been really applied to the new updated model.

<br>

The fact that it deletes materials came up recently in a PR here: [https://github.com/ynput/ayon-blender/pull/275/changes#diff-a0090cb17413e24bb2fddf8604f1c590fbde46fa62739f2466fc9462a6c77dbfR203-R205](https://github.com/ynput/ayon-blender/pull/275/changes#diff-a0090cb17413e24bb2fddf8604f1c590fbde46fa62739f2466fc9462a6c77dbfR203-R205) 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1491039808133533927)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0641]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=930a2a50327711f1b13f05742f042725)

moonyuet:
## Changelog Description
This PR is to make sure cache loader not removing material if there are any other objects linking to that during update
Fixes: https://github.com/ynput/ayon-blender/issues/282

n/a

## Testing notes:
1. Load Cache with material or assign some material to the cache object
2. Make sure the assigned material also assigned to another objects(e.g. object B)
3. Updating the loaded cache asset
4. The material still preserves in the another objects(e.g. object B)

### DIFF
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -207,6 +207,10 @@ class CacheModelLoader(plugin.BlenderLoader):
                     material = material_slot.material
                     if not material:
                         continue
+
+                    if material.users > 1:
+                        continue
+
                     bpy.data.materials.remove(material)
                 bpy.data.meshes.remove(obj.data)
             elif obj.type == 'EMPTY':
--- a/client/ayon_blender/plugins/load/load_cache.py
+++ b/client/ayon_blender/plugins/load/load_cache.py
@@ -200,18 +200,6 @@ class CacheModelLoader(plugin.BlenderLoader):
 
         for obj in objects:
             if obj.type == 'MESH':
-                # QUESTION: Are we sure we want to remove the materials when
-                #  removing the mesh? What if the material is shared with
-                #  other objects?
-                for material_slot in list(obj.material_slots):
-                    material = material_slot.material
-                    if not material:
-                        continue
-
-                    if material.users > 1:
-                        continue
-
-                    bpy.data.materials.remove(material)
                 bpy.data.meshes.remove(obj.data)
             elif obj.type == 'EMPTY':
                 objects.extend(obj.children)


LiborBatek (approved):
All been working well and without any glitches...

LGTM

client/ayon_blender/plugins/load/load_cache.py
```diff
@@ -207,6 +207,10 @@ def _remove(self, asset_group):
                     material = material_slot.material
                     if not material:
                         continue
+
+                    if material.users > 1:
+                        continue
```
BigRoy: I'm not entirely convinced this is the right way.

There are just too many caveats:
- The user may have purposefully assigned a material to the object that did not come from the loaded cache - when removing the loaded objects I feel it's wrong that we're also deleting that material.
- When loading a cache where multiple objects share the same material, the user count would be larger than 1 but still zero after removal/update. But we're failing to remove those materials because we've found more than one object using it.

I think the case we're looking for is that if we're loading, we store which materials were loaded - and then if we were to remove the loaded cache, we delete only those that we KNOW came from loading it initially.

Also, if I load a cache - and assign a linked material (e.g. published look) then doesn't that error here when trying to remove the material because it can't remove what's linked, or?

Anyway, my point of view is that we must either:
- Not delete it
- Or must be extremely confident that what we're deleting is the right thing to delete.

@LiborBatek if you were to load a cache, assign a material yourself... when updating or removing the cache would you also expect that material you had created to persist? if so, then this logic here feels wrong.

I wonder if we're not better off just not touching the materials 🤔 OR explicitly storing which got loaded in so that if we update/remove that we only ever touch those materials. (just not sure how reliably we can store such a link in Blender)

BigRoy (changes requested):
With this condition added we can better remove full material deletion code. Because the material shou

client/ayon_blender/plugins/load/load_cache.py
BigRoy: Client confirmed they assigned their own materials, and do not want them deleted whatsoever.

As such, I think the right way for now is to delete the code for removing of materials and keep them around.

LiborBatek: not sure whats the point here...as it works exactly like the client expects...no material got deleted when testing, also not sure what are we talking about when refering `pointcache` products never having any materials assigned when just loaded...so all shading happens using materials either from the current workfile or just created/appended from elsewhere 

or did I miss something?

BigRoy: > or did I miss something?

Loading a USD cache _may_ introduce materials from that load. ;)

Als the current logic may still sometimes delete materials - and my point is, we shouldn't really.

moonyuet: > Als the current logic may still sometimes delete materials - and my point is, we shouldn't really.

would it be safer to instead checking users is equal or more than 1?

BigRoy: Not really - if the material is on this object, it _should_ always have at least one user - and at that point the check becomes completely useless. Hence the reasoning to just remove the logic altogether.

moonyuet: So maybe the best case scenario should be removing the code which does the material deletion then

BigRoy: Exactly,  yes.

[> As such, I think the right way for now is to delete the code for removing of materials and keep them around.](https://github.com/ynput/ayon-blender/pull/283#discussion_r3051178498)

moonyuet: > Exactly, yes.
> 
> [> As such, I think the right way for now is to delete the code for removing of materials and keep them around.](https://github.com/ynput/ayon-blender/pull/283#discussion_r3051178498)

Committed to remove the logic: [eca62c5](https://github.com/ynput/ayon-blender/pull/283/commits/eca62c52a0ec2f12d18ee97f9d584b546389fcd9)

BigRoy (approved):
Looks about right - we may have some lingering materials in some cases, but it's less worse than the previous behavior.

LiborBatek (approved):
All been working well for both `4.5.x` and `5.x` Blender.

LGTM

PR #287, opened 2026-04-08
Blender 5.0: Media type setting set to the correct attribute for multilayer exr setting

### LINKED ISSUE #285: Media type error in Blender 5.0

## Current Behavior
When you are creating render with `ayon+settings://blender/RenderSettings/multilayer_exr` on, and then delete render instance, and then turn off the   `ayon+settings://blender/RenderSettings/multilayer_exr`, and re-create the render instance. It would error out as followed screenshot. And this bug only happened in blender 5.x.

Image

## Expected Behavior
It shouldn't be errored out.

### Steps To Reproduce:

1.  Create render with `ayon+settings://blender/RenderSettings/multilayer_exr` on
2. Delete render instance
3. Turn off  `ayon+settings://blender/RenderSettings/multilayer_exr` 
4. Change image format and re-create render instance

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

moonyuet:
## Changelog Description
This PR is to ensure media type setting set to the correct attributes so that it won't hit the enum error as stated in https://github.com/ynput/ayon-blender/issues/285

Put it into draft first and no guarantee there is some issue hit when passing the attributes to collector.

## Testing notes:
1. Create render instance by rendersetup and make sure `ayon+settings://blender/RenderSettings/multilayer_exr` on and image format is `OpenEXR`
2. Publish should be successful
3. Create render instance by rendersetup again but this time  `ayon+settings://blender/RenderSettings/multilayer_exr` off and image format is still `OpenEXR`.
4. Publish should be also successful

### DIFF
--- a/client/ayon_blender/api/render_lib.py
+++ b/client/ayon_blender/api/render_lib.py
@@ -58,8 +58,11 @@ def set_render_format(ext: str, multilayer: bool):
     bpy.context.scene.render.use_file_extension = True
     image_settings = bpy.context.scene.render.image_settings
 
-    if multilayer and lib.get_blender_version() >= (5, 0, 0):
-        image_settings.media_type = "MULTI_LAYER_IMAGE"
+    if lib.get_blender_version() >= (5, 0, 0):
+        if multilayer:
+            image_settings.media_type = "MULTI_LAYER_IMAGE"
+        else:
+            image_settings.media_type = "IMAGE"
 
     if ext == "exr":
         file_format = "OPEN_EXR_MULTILAYER" if multilayer else "OPEN_EXR"


BigRoy (approved):
Code LGTM.

Should we also double check whether the behavior for Collect Render still works reliably, because it does concern [the file type but does not check the media type](https://github.com/ynput/ayon-blender/blob/78dd5cd3d3ac2accbec6fd789cf62984f40f042a/client/ayon_blender/plugins/publish/collect_render.py#L231)? Should it check it? Or is the file type sufficient there?

moonyuet:
> Code LGTM.
> 
> Should we also double check whether the behavior for Collect Render still works reliably, because it does concern [the file type but does not check the media type](https://github.com/ynput/ayon-blender/blob/78dd5cd3d3ac2accbec6fd789cf62984f40f042a/client/ayon_blender/plugins/publish/collect_render.py#L231)? Should it check it? Or is the file type sufficient there?

I believe it should check the media type correctly, as far as I checked on the code and test.  But I would let @LiborBatek to do some stress test and see if any other edge case with that new feature of `media type`

LiborBatek (approved):
I was able to produce a publish and rendered sequence and also in different circumstances (single layer and multilayered exr).

Including succesful integration of the render on Ayon.

LGTM now

PR #289, opened 2026-04-12
Fix UI refresh interval on linux

Plyrolith:
## Changelog Description
Linux Qt UI was very sluggish. The app handler refresh interval was set too large. This was probably set up for an issue on MacOS and unnecessarily applied to Linux (probably never tested). Flipping a switch took multiple seconds since animation playback was so slow.

Changed the global variable `TIMER_INTERVAL` to be 0.01 with Darwin to be the exception at 0.1. The "normal" interval was the exception for Windows instead.

## Testing notes:
Tested and verified on these Linux system configurations:
- AlmaLinux9, NVIDIA, Xorg, Blender 4.5 + 5.1
- Fedora 43, NVIDIA, Wayland, Blender 4.5 + 5.1
- Fedora 43, Intel Arc, Wayland, Blender 4.5 + 5.1

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -38,7 +38,7 @@ PREVIEW_COLLECTIONS: Dict = dict()
 # This seems like a good value to keep the Qt app responsive and doesn't slow
 # down Blender. At least on macOS I the interface of Blender gets very laggy if
 # you make it smaller.
-TIMER_INTERVAL: float = 0.01 if platform.system() == "Windows" else 0.1
+TIMER_INTERVAL: float = 0.1 if platform.system() == "Darwin" else 0.01
 
 
 def execute_function_in_main_thread(f):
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
LGTM.

@rdelillo can you confirm the lower value of `0.01` is still problematic on Mac OS?

rdelillo:
> LGTM.
> 
> @rdelillo can you confirm the lower value of `0.01` is still problematic on Mac OS?

@BigRoy Doesn't look like it anymore. This patch is more than 2 years old.
It all works fine for me on Blender 5.1 with `0.01`. Do we take the opportunity to set it back for all paltforms ?

BigRoy:
> > LGTM.
> > @rdelillo can you confirm the lower value of `0.01` is still problematic on Mac OS?
> 
> @BigRoy Doesn't look like it anymore. This patch is more than 2 years old. It all works fine for me on Blender 5.1 with `0.01`. Do we take the opportunity to set it back for all paltforms ?

I think Blender 4.2 LTS and Blender 4.5 LTS are still good candidates to double check perhaps.

rdelillo:
Confirmed it works great on 4.2LTS and 4.5LTS on macOS without the special treatment.
Thanks a lot @Plyrolith for this !

PR #290, opened 2026-04-14
Change PySide install hook to use target in local dir

Plyrolith:
## Changelog Description
The current implementation to add Qt bindings to Blender is highly problematic and breaks established standards.

### Issues

- Security: User will be asked to run arbitrary Python code with elevated privileges. This should **never** happen under any circumstance. Most users and admins will not audit the code and underlying dependency packages are not even audited by developers in general.
- Deployment: Even if the security flaw above is avoided, admins will have to deploy a pre-patched Blender. This increases workload and complexity. Official Blender package sources cannot be used directly.
- Compatibility: The patched Blender installation is set to the specific package version which might cause problems for other extensions.
- Stability: Blender's system components are altered, which might generally cause unexpected behavior and even crashes.

The new install routine utilizes `pip install --target` to install Qt bindings into `AYON_LAUNCHER_LOCAL_DIR / application_packages / blender-pyX.X`. This directory is guaranteed to be in user space and will not conflict with system resources.

### Overview

- The hook will generate the path to a subdir of `application_packages` based on Blender's Python version.
- Blender's packages are added to `PYTHONPATH`.
- If the path exists, it is added to `PYTHONPATH` as well.
- The hook checks if Qt bindings are available.
- If they are not available, it will create the subdir and install PySide using `pip install --target` with Blender's Python executable.
- If it didn't exist before, the path is now prepended to `PYTHONPATH`.
- Blender is started with the extended environment.


moonyuet:
@BigRoy just wanna know your opinion on this? Generally this makes sense to me especially when the installation of pyside6 probably need admin rights if the artist requires to install the library. 

I can imagine the code can be simplified as the implementation in the hook and virtualvenv class or any useful functions can be put as scripts into api folder.

BigRoy:
Switching to a safer mode of install then forcing it into Blender itself I think is a good thing - however, us managing anything at all should still be optional.

So:
- Hook setting should remain, this should be allowed to be disabled so that IT can decide to host dependencies completely on their own behalf.
- Not sure about the usage of `os.environ.get("AYON_LAUNCHER_LOCAL_DIR", "")`. It should likely not fall back to an empty string? @iLLiCiTiT any concerns with using this env var in particular to store DCC-specific additional files like a venv install on prelaunch hook.
- ~~I think the venv init should not be enforced before checking whether the dependency is available. It should still only deploy if studio customizations (e.g. custom IT) didn't make the dependency available out of the box or in some other manner.~~ This actually does seem implemented by first doing the import check.
- I don't like the run on import behavior of `ensure_qt_binding`. Rather have an explicit call to to a function somewhere on our logic.
- If for whatever reason our now runtime logic triggers later than anything else requiring Qt imports (e.g. pipeline-specific launch scripts that a studio may configure) then the Qt dependency now comes too late, instead of being available directly on launch. This may be an unwanted side-effect.
- I think the fact that this is a `.venv` is slightly confusing. To my knowledge a `venv` is intended to be an isolated environment on its own, and now we're stacking a venv against Blender's environment - which I know is reasonable but it's something to be aware of. It looks like we're "freezing" Blender's env - and then installing our dependency against it, which to my knowledge will still 'install' the frozen requirements if lacking in the venv itself which may take up more disk-space? (But not sure if I'm right here.)

So even though I think it offers some good ideas I don't think it's fully there to be merged. We'll need to discuss internally how much we like this change further, because conceptually some of it may transfer to other DCCs that do similar things. @antirotor @iLLiCiTiT @philippe-ynput ?

Plyrolith:
Here's some input from my side, for what it's worth:

https://github.com/ynput/ayon-blender/pull/290#issuecomment-4336063694

>     * Hook setting should remain

- Since Qt needs to be available to Blender (and not the launcher) it was easiest to perform checks & installs directly from within.
- I also thought it's preferable to just install the venv, compared to a plain crash if Qt is unavailable.
- Any IT department that requires more specific bindings will likely already have a them in place, which is checked.
- Again, the install is in user-space (just like AYON plugin updates for example) and low impact. I didn't see any downside to it.

I can see how you prefer control though, so the implementation as a hook is understandable.

>     * Not sure about the usage of `os.environ.get("AYON_LAUNCHER_LOCAL_DIR", "")

Analyzing the launcher code I assumed this was the default location to add required data.
https://github.com/ynput/ayon-launcher/blob/f8c4f73e1ba846a0020cd896e76260f44cef388e/common/ayon_common/utils.py#L129

The launcher always sets this variable, even if not present in the environment. Seemed safe to me. Empty string default is just for type linting.

>     * ~I think the venv init should not be enforced before checking whether the dependency is available.

It isn't. If Qt is available on import, the venv won't even initialize.

>     * I don't like the run on import behavior of `ensure_qt_binding`.

It could be wrapped in another function. This will add more linting exceptions though, because we need to *import / run / import*, since the binding needs to be ensured before `ayon_core` is imported. Importing it with missing bindings will cause a blocking exception.

>     * [...] runtime logic triggers later than anything else requiring Qt

The logic triggers on Blender startup, separate from AYON launcher. Templates are initiated even before add-ons, so it's very early in the Blender init. If another binding is already provided, it will use that one instead. I see no issue with the timing.

>     * [...] to my knowledge will still 'install' the frozen requirements if lacking in the venv itself

I looked into this - the `EnvBuilder` can be initialized using `system_site_packages=True`, which will enable it to reuse existing packages (unless there's a version requirement mismatch, of course). Thanks for making me aware of this, I will change this across other projects. :)

I propose to have one more go at it, with the following changes:

- Restore the hook (and its settings).
- Remove the old system install methods from the hook.
- Build the venv logic directly into the hook.
- Use a reduced VenvManager class in the hook; only required methods and properties
- Ensure the venv is built with `system_site_packages=True`

Since the hook needs to call Blender's Python executable, I need to perform the `check / init / check / install` logic using subprocess, but it should work.

We still need to initialize the env within Blender on runtime though, so the logic has to be split:
- Hook: `check / init / check / install`
- Blender startup: `check / init`

I will need to figure out how to make sure the logic is available in both places without redundancy.

Let me know what you think.

BigRoy:
> We still need to initialize the env within Blender on runtime though, so the logic has to be split:
> 
> * Hook: `check / init / check / install`
> * Blender startup: `check / init`

I don't think that's the case. You'd just do:

- confirm it's already available
- if not, generate the venv
- then we should just be able to add the venv to launch context env `PYTHONPATH`?

Anyway, if we move it to a hook - it also saves us from having to need to trigger this in Blender and hence solve the need for  that run on import issue. If we can get the hook be close to current performance that'd be amazing. (I'm fine if the installation is slightly slower, but runs after should be fast as well. Which makes me think that first check might need to include the venv already if it exists.)

> The launcher always sets this variable, even if not present in the environment. Seemed safe to me. Empty string default is just for type linting.

In that case just use `os.environ["AYON_LAUNCHER_LOCAL_DIR"]`. If you're expecting it to be set ALWAYS. Otherwise it'd just hide potentially dangerous issues.

client/ayon_blender/venvman.py
```diff
@@ -0,0 +1,375 @@
...
+            cwd=str(bin_path),
+        )
+
+        thread_out = Thread(target=self.reader, args=(proc.stdout,))
+        thread_out.start()
+        thread_err = Thread(target=self.reader, args=(proc.stderr,))
+        thread_err.start()
+        proc.wait()
+        thread_out.join()
+        thread_err.join()
```
iLLiCiTiT: Looks like all you need it this?
```suggestion
        proc = subprocess.run(
            (context.env_exe, fn),
            stdout=sys.stderr,
            stderr=sys.stderr,
            cwd=str(bin_path),
        )
```

client/ayon_blender/venvman.py
```diff
@@ -0,0 +1,375 @@
...
+            (self.pip_path, "freeze"),
+            capture_output=True,
+            text=True,
+        ).stdout
+        freeze_packages = freeze.split("\n")
+
+        # Read the current requirements and build a list of package names
+        with open(self.requirements_file, "r") as req_file:
+            req_lines = req_file.readlines()
+            packages = [line.split("==")[0] for line in req_lines if line]
```
iLLiCiTiT:
```suggestion
        packages = [line.split("==")[0] for line in req_lines if line]
```

client/ayon_blender/qt_binding.py
```diff
@@ -0,0 +1,69 @@
...
+    print(f"Attempting {qt_binding} import from package libraries...")
+    if check_import(qt_binding):
+        print(f"Qt binding {qt_binding} available in package libraries.")
+        return
+
+    # Prepare the virtual environment
+    venv_path = Path(
+        os.environ.get("AYON_LAUNCHER_LOCAL_DIR", ""),
+        "virtualenvs",
+        f"blender-{bpy.app.version_string}",
```
iLLiCiTiT: This should use `get_launcher_local_dir` from `ayon_core.lib`.
```suggestion
        get_launcher_local_dir(
            "virtualenvs",
            f"blender-{bpy.app.version_string}",
        ),
```

client/ayon_blender/venvman.py
```diff
@@ -0,0 +1,375 @@
...
+from pathlib import Path
+import subprocess
+import sys
+from threading import Thread
+from urllib.parse import urlparse
+from urllib.request import urlretrieve
+from venv import EnvBuilder
+
+
+class VenvManager(EnvBuilder):
```
iLLiCiTiT: Would this work for windows, linux and mac?

client/ayon_blender/venvman.py
```diff
@@ -0,0 +1,375 @@
+from __future__ import annotations
```
iLLiCiTiT: Does it make sense to separate it from qt_binding? If yes could it be renamed to `venv_manager.py`?

iLLiCiTiT:
Does it have to be venv? You can skip venv creation and pip install to a directory `pip install --target <dirpath> PySide6`. Technically the same thing but without venv "sh*t", just add the path to PYTHONPATH in launch hook (or in `sys.path` inside blender).

I would consider that the directory where it is being installed is a hash of the executable path. Or, the opposite, if we're sure it is save we can skip blender version and use python version instead -> `blender-py{python version}`, so if multiple blenders share the same python version they can re-use the dependencies.

It should be possible to control it using settings, in case something goes wrong, this should not block artists from work. It is "red button" option to allow artists to work even if pipeline is broken.

Plyrolith:
> Does it have to be venv? You can skip venv creation and pip install to a directory `pip install --target <dirpath> PySide6`. Technically the same thing but without venv "sh*t", just add the path to PYTHONPATH in launch hook (or in `sys.path` inside blender).
> 
> I would consider that the directory where it is being installed is a hash of the executable path. Or, the opposite, if we're sure it is save we can skip blender version and use python version instead -> `blender-py{python version}`, so if multiple blenders share the same python version they can re-use the dependencies.
> 
> It should be possible to control it using settings, in case something goes wrong, this should not block artists from work. It is "red button" option to allow artists to work even if pipeline is broken.

This should work and makes more sense, since all you need is Qt. The only downside would be that it definitely will install all dependencies, even if they're already available in the main interpreter's site-packages. Probably still more lightweight than a full venv.

iLLiCiTiT:
> The only downside would be that it definitely will install all dependencies, even if they're already available in the main interpreter's site-packages.

No, it does install only what is not in site-packages (we use it for runtime dependencies).

BigRoy:
> This should work and makes more sense, since all you need is Qt. The only downside would be that it definitely will install all dependencies, even if they're already available in the main interpreter's site-packages. Probably still more lightweight than a full venv.

Pretty sure pip install with target dir does not do that - it only does so for the missing dependencies. 🤔

Plyrolith:
Re-implemented with the existing hook, using `pip install --target` and minimal changes. Location is `AYON_LAUNCHER_LOCAL_DIR / program_packages` (seemed in-line with existing naming) and `blender-pyX.X` as suggested.

client/ayon_blender/hooks/pre_pyside_install.py
```diff
@@ -138,107 +136,71 @@ def inner_execute(self):
...
+        python_minor = self.get_python_version(python_executable)
+        local_dir = get_launcher_local_dir()
+        app_packages = os.path.join(
+            local_dir, "application_packages", f"blender-py{python_minor}"
+        )
+
+        # Prepend application packages dir to Python path and check availability again
+        if os.path.exists(app_packages):
+            self.prepend_to_pythonpath([app_packages])
+            if self.is_pyside_installed(python_executable, qt_binding):
```
BigRoy: We now check `is_pyside_installed` twice which seems redundant.
I would:

- prepend the site-packages.
- prepend the relevant app_packages path (if exists)
- test if pyside is available
- if not, then install to app packages folder?
- if did not prepend the app packages folder previously (due to not existed), then prepend it?

```
app_packages_exists = os.path.exists(app_packages)
if app_packages_exists:
    python_paths.append(app_packages)

self.prepend_to_pythonpath(python_paths)
if self.is_pyside_installed(...):
    return

# We will need to install it
if not app_packages_exists:
    os.makedirs(app_packages, exists_ok=True)
    self.prepend_to_pythonpath([app_packages])
result = self.install_pyside(...)
```

Plyrolith: My reason for the double check is to leave PYTHONPATH untouched in case the Blender app natively comes with Qt bindings preinstalled. Only after the first “raw” check fails, we prepend the “applications_packages”.
E.g. if a studio chooses to ship Blender with Qt included AFTER it had been started without it, it’ll prefer the included one even if the hook had already installed PySide before.

BigRoy: > E.g. if a studio chooses to ship Blender with Qt included AFTER it had been started without it, it’ll prefer the included one even if the hook had already installed PySide before.

In that case, just make sure to slip in the path from Blender itself before ours. That way, if both exist and provide it, still the first one will be used.

Plyrolith: > > E.g. if a studio chooses to ship Blender with Qt included AFTER it had been started without it, it’ll prefer the included one even if the hook had already installed PySide before.
> 
> In that case, just make sure to slip in the path from Blender itself before ours. That way, if both exist and provide it, still the first one will be used.

Updated to perform only one availability check.
- If `application_packages` for the Blender version exists, it'll be appended after internal packages.
- `is_pyside_installed` is run.
- If no Qt bindings are available, PySide will be installed.
- If `application_packages` did not exist before, it will now be prepended to `PYTHONPATH`. It will only be on position 0 directly after install. Consecutive runs will place it behind Blender's internal packages.

BigRoy:
@Plyrolith looks good to me - can you update the PR description to describe how it works now? Since venvmanager etc is not there anymore.

BigRoy:
@LiborBatek @moonyuet can you test this?

- Make sure to remove already installed PySide6 from the Blender install to ensure it re-installs.
- Might be good to try and run via CLI to actually see the logs of the pre-launch hooks.

@rdelillo can you test this on MacOS?

moonyuet (approved):
Tested successfully in blender 5.0

client/ayon_blender/hooks/pre_pyside_install.py
```diff
@@ -133,112 +124,81 @@ def inner_execute(self):
             )
             return
 
-        # Check if PySide2 is installed and skip if yes
+        # Check if application packages dir exists for Blender's Python version
+        python_minor = self.get_python_version(python_executable)
+        local_dir = get_launcher_local_dir()
+        app_packages = os.path.join(
+            local_dir, "application_packages", f"blender-py{python_minor}"
+        )
```
rdelillo, line 132: picky, I'd personally suggest a small tweak and use 
```python
from ayon_core import lib
blender_target_path = lib.get_addons_resources_dir(
    "ayon_blender",
    f"blender-py{python_minor}-pyside",
)
```

rdelillo (approved):
Confirm it works as expected on MacOs.

client/ayon_blender/hooks/pre_pyside_install.py
Plyrolith: The thought was to have one unified dir in case you might need to use a similar hook in other applications. Application name is already specified in the sub dir. I tried to keep it consistent with "dependency_packages".

BigRoy: @iLLiCiTiT any reason to choose for either `get_addons_resources_dir` or `get_launcher_local_dir` from your perspective?

Plyrolith: Looking at “get_addons_resources_dir” in ayon_core this actually makes more sense and seems to be made for this kind of data. I think addon resources makes more sense. Let me know if I should commit.

iLLiCiTiT: `get_addons_resources_dir` uses `get_launcher_storage_dir` by default which can be shared directory for multiple machines, I don't think it is good idea.

iLLiCiTiT, line 132:
```suggestion
        app_packages = get_launcher_local_dir(
            "application_packages", f"blender-py{python_minor}"
        )
```

iLLiCiTiT: This should contain full version, there might be differences in patch versions.

Shouldn't this use hash of the python path instead? Safer in case you have 2 blender versions with same python version but different venv.

```python
import hashlib

def get_path_hash(path: str) -> str:
    return hashlib.sha256(
        path.encode("utf-8")
    ).hexdigest()[:8]


....
        short_hash = get_path_hash(python_path)
        app_packages = get_launcher_local_dir(
            "application_packages", f"blender-py{short_hash}"
        )

```

client/ayon_blender/hooks/pre_pyside_install.py
```diff
@@ -133,112 +124,81 @@ def inner_execute(self):
...
+
+        # Change PYTHONPATH to contain blender's packages as first
+        python_paths = [
+            python_lib,
+            os.path.join(python_lib, "site-packages"),
+        ]
+
+        # Append application packages dir to Python path, if it exists
+        if os.path.exists(app_packages):
+            python_paths.append(app_packages)
```
iLLiCiTiT, line 142: I don't understand why not just add it here? If it doesn't exist it simply won't be used, at the end it is always added.
```suggestion
            app_packages,
        ]
```

client/ayon_blender/hooks/pre_pyside_install.py
```diff
@@ -133,112 +124,81 @@ def inner_execute(self):
...
-            )
+        # Install PySide2 into application packages
+        result = self.install_pyside(
+            python_executable,
+            qt_binding,
+            qt_binding_version,
+            app_packages,
+        )
+        if app_packages not in python_paths:
+            self.prepend_to_pythonpath([app_packages])
```
iLLiCiTiT, line 158:
```suggestion
```
Suggestion related to https://github.com/ynput/ayon-blender/pull/290/changes#r3274024297

client/ayon_blender/hooks/pre_pyside_install.py
BigRoy: Just because then we'll still be force adding it into the launched executable's env paths, even if it does not exist and is perhaps not needed. Which seemed not pretty.

iLLiCiTiT: Is it because of `if self.is_pyside_installed(python_executable, qt_binding):`? Why we can't just run it? Why we're adding site-packages of the python we're starting to the pythonpath, shouldn't that be done by the python itself?

Shouldn't we also run the `is_pyside_installed` before we do add `app_packages`, even if `app_packages` exists it still might not be needed...

PR #292, opened 2026-04-20
Get AYON property data return dictionary data type 

### LINKED ISSUE #293: Invalid property data type causes Blender crash

## Current Behavior
If users are using old workfile with avalon data and open publisher UI(or other plugins), and it would crash Blender due to the caching instance not being able to convert the data type from blender to dictionary.
See https://github.com/ynput/ayon-core/pull/1787#issuecomment-4268406335 
## Expected Behavior
When users open old workfile with avalon data, it would not crash Blender anymore.

### Steps To Reproduce:

1. Open old workfile with avalon data
2. Open Publisher UI
3. It then crashes Blender

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

moonyuet:
## Changelog Description
This PR is to fix the blender crash when ayon property data return irrelevant data type which is not readable for python.
Fixes: https://github.com/ynput/ayon-blender/issues/293
Tested with load placeholder, load blendscene, load blend with new ayon data.
Tested with old avalon data in the publisher UI and placeholder and loaders as mentioned above
Tested in Blender 4.x and 5.x 

## Testing notes:
1. Use old workfile with avalon data
2. Open publisher or any UI
3. Should not crash blender and show up all the data in Publisher
4. Use workfile with ayon data
5. Should be also not crashing the publisher or any UI

### DIFF
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -519,7 +519,7 @@ def _discover_gui() -> Optional[Callable]:
     return None
 
 
-def get_ayon_property(node):
+def get_ayon_property(node) -> Dict:
     property = node.get(AYON_PROPERTY)
     if not property:
         # Backwards compatibility: Update legacy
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Been working in all circumstances in the testing steps.

LGTM

PR #295, opened 2026-04-29
Extract USD: Synchronize USD export options from Ayon to Publisher

### LINKED ISSUE #291: Ayon Blender USD doesnt export animation

## Current Behavior
When publishing an USD from blender, it does not export the animation

I think whats happening is that it copies the default USD export settings from blender, but the default for animation is turned of. So only static USDs can be published. Exposing the checkbox in the publisher would fix this issue

### Steps To Reproduce:

Publish any animated mesh as a USD using the Ayon publisher

Windows

### Relevant log output:

```shell

```

moonyuet:
## Changelog Description
This PR is to synchronize USD export options from Ayon to Publisher, and to allow user to choose their preferred options.
Resolve https://github.com/ynput/ayon-blender/issues/291

Build and install addon with this branch

## Testing notes:
1. Go to `ayon+settings://blender/publish/ExtractUSD` 
image

2. Create USD instance
3. You can see the options in the publisher UI
4. Publish

### DIFF
--- a/client/ayon_blender/plugins/publish/extract_usd.py
+++ b/client/ayon_blender/plugins/publish/extract_usd.py
@@ -26,7 +26,7 @@ class ExtractUSD(plugin.BlenderExtractor,
     export_mesh_colors = True
     use_instancing = True
 
-    overrides = []
+    overrides: list[str] = []
 
     def process(self, instance):
         if not self.is_active(instance.data):
--- a/client/ayon_blender/plugins/publish/extract_usd.py
+++ b/client/ayon_blender/plugins/publish/extract_usd.py
@@ -1,6 +1,4 @@
 import os
-from collections import OrderedDict
-
 import bpy
 
 from ayon_core.pipeline import KnownPublishError, OptionalPyblishPluginMixin
@@ -154,7 +152,7 @@ class ExtractUSD(plugin.BlenderExtractor,
         if not overrides:
             return defs
 
-        override_defs = OrderedDict({
+        override_defs = {
             "convert_orientation": BoolDef("convert_orientation",
                     label="Convert Orientation",
                     tooltip="Convert orientation axis to a different"
@@ -205,8 +203,7 @@ class ExtractUSD(plugin.BlenderExtractor,
                 label="Instancing",
                 tooltip="Whether to use USD instancing for duplicated objects or not.",
                 default=cls.use_instancing),
-
-        })
+        }
 
         for key, value in override_defs.items():
             if key not in overrides and key not in {"forward_axis", "up_axis"}:
--- a/client/ayon_blender/plugins/publish/extract_usd.py
+++ b/client/ayon_blender/plugins/publish/extract_usd.py
@@ -150,56 +150,73 @@ class ExtractUSD(plugin.BlenderExtractor,
             return defs
 
         override_defs = {
-            "convert_orientation": BoolDef("convert_orientation",
-                    label="Convert Orientation",
-                    tooltip="Convert orientation axis to a different"
-                    " convention to match other applications.",
-                    default=cls.convert_orientation),
-            "forward_axis": EnumDef("forward_axis",
-                    label="Forward Axis",
-                    items=orientation_axes,
-                    default="Z",
-                    visible=visible),
-            "up_axis": EnumDef("up_axis",
-                    label="Up Axis",
-                    items=orientation_axes,
-                    default="Y",
-                    visible=visible),
+            "convert_orientation": BoolDef(
+                "convert_orientation",
+                label="Convert Orientation",
+                tooltip="Convert orientation axis to a different convention"
+                        " to match other applications.",
+                default=cls.convert_orientation,
+            ),
+            "forward_axis": EnumDef(
+                "forward_axis",
+                label="Forward Axis",
+                items=orientation_axes,
+                default="Z",
+                tooltip="Forward Axis for orientation conversion.",
+                visible=visible,
+            ),
+            "up_axis": EnumDef(
+                "up_axis",
+                label="Up Axis",
+                items=orientation_axes,
+                default="Y",
+                tooltip="Up Axis for orientation conversion.",
+                visible=visible,
+            ),
             "export_animation": BoolDef(
                 "export_animation",
                 label="Export Animation",
                 tooltip="Whether to export animation data or not.",
-                default=cls.export_animation),
+                default=cls.export_animation,
+            ),
             "export_hair": BoolDef(
                 "export_hair",
                 label="Export Hair",
                 tooltip="Whether to export hair/fur systems or not.",
-                default=cls.export_hair),
+                default=cls.export_hair,
+            ),
             "export_uvmaps": BoolDef(
                 "export_uvmaps",
                 label="Export UV Maps",
                 tooltip="Whether to export UV map data or not.",
-                default=cls.export_uvmaps),
+                default=cls.export_uvmaps,
+            ),
             "export_normals": BoolDef(
                 "export_normals",
                 label="Export Normals",
                 tooltip="Whether to export normal data or not.",
-                default=cls.export_normals),
+                default=cls.export_normals,
+            ),
             "export_materials": BoolDef(
                 "export_materials",
                 label="Export Materials",
-                tooltip="Whether to export material assignments and data or not.",
-                default=cls.export_materials),
+                tooltip="Whether to export material assignments and data or"
+                        " not.",
+                default=cls.export_materials,
+            ),
             "export_mesh_colors": BoolDef(
                 "export_mesh_colors",
                 label="Export Mesh Colors",
                 tooltip="Whether to export mesh color data or not.",
-                default=cls.export_mesh_colors),
+                default=cls.export_mesh_colors,
+            ),
             "use_instancing": BoolDef(
                 "use_instancing",
                 label="Instancing",
-                tooltip="Whether to use USD instancing for duplicated objects or not.",
-                default=cls.use_instancing),
+                tooltip="Whether to use USD instancing for duplicated objects"
+                        " or not.",
+                default=cls.use_instancing,
+            ),
         }
 
         for key, value in override_defs.items():
--- a/client/ayon_blender/plugins/publish/extract_usd.py
+++ b/client/ayon_blender/plugins/publish/extract_usd.py
@@ -15,7 +15,14 @@ class ExtractUSD(plugin.BlenderExtractor,
     hosts = ["blender"]
     families = ["usd"]
 
+    # Settings
     convert_orientation = False
+    export_animation=False
+    export_hair=False
+    export_uvmaps=True
+    export_normals=True
+    export_materials=True
+    use_instancing=True
 
     def process(self, instance):
         if not self.is_active(instance.data):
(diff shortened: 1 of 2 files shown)


BigRoy:
This will now flood the publisher UI with more attributes to set - which will end up being more over time (as blender introduces new arguments, etc.)

I think we should do similar to Maya's Alembic options where the admin can define the Exposed overrides from settings to allow limiting the attributes that are available?

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -165,6 +160,11 @@ def get_attr_defs_for_instance(cls, create_context, instance):
             return defs
 
         override_defs = OrderedDict({
+            "convert_orientation": BoolDef("convert_orientation",
+                    label="Convert Orientation",
+                    tooltip="Convert orientation axis to a different"
+                    " convention to match other applications.",
+                    default=cls.convert_orientation),
```
moonyuet: @BigRoy should we exclude this as it is not working when we disable this options?
I reverted the change already btw.

BigRoy: Not working in what way? Models come through wrong? or just doesn't export at all?

I think we should expose the overrides that we had previously (so same defaults as before)... and if this convert orientations does work then it's still relevant for settings.

moonyuet: > Not working in what way? Models come through wrong? or just doesn't export at all?
> 
> I think we should expose the overrides that we had previously (so same defaults as before)... and if this convert orientations does work then it's still relevant for settings.

It doesn't synchronize for some reasons. Let me try again and double confirm. It could be some settings messing up  between dev and general mode

BigRoy: I think maybe you'd need to also set some other arguments to define what to convert to? Maybe `export_global_forward_selection ` and `export_global_up_selection` perhaps?

moonyuet: > I think maybe you'd need to also set some other arguments to define what to convert to? Maybe `export_global_forward_selection ` and `export_global_up_selection` perhaps?

Fixed in [0105932](https://github.com/ynput/ayon-blender/pull/295/commits/01059323146a10c35980bbe09eb96bd7b648d320)

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -114,47 +120,95 @@ def process(self, instance):
...
+            "use_instancing": BoolDef(
+                "use_instancing",
+                label="Instancing",
+                tooltip="Whether to use USD instancing for duplicated objects or not.",
+                default=cls.use_instancing),
+
+        })
+
+        for key, value in override_defs.items():
+            if key not in overrides and key not in {"forward_axis", "up_axis"}:
```
BigRoy, line 223: What makes `forward_axis` and `up_axis` special? this makes it look like, if there's at least one exposed override - then show these (which seems wrong?). Either they always show, OR they show if certain other attributes are present. But I feel like they shouldn't suddenly show if only I have `export_materials` enabled or `use_instancing`?

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -84,21 +99,12 @@ def process(self, instance):
...
                 filepath=filepath,
-                root_prim_path="",  
+                root_prim_path="",
                 selected_objects_only=True,
                 relative_paths=False,
-                export_animation=False,
-                export_hair=False,
-                export_uvmaps=True,
-                # TODO: add for new version of Blender (4+?)
-                # export_mesh_colors=True,
```
BigRoy, line 97: Can we add the `export_mesh_colors` as well since we're at it?

client/ayon_blender/plugins/publish/extract_usd.py
moonyuet: These attributes are only shown when `convert orientation` enabled (in either Ayon or overrides setting), you can notice the visible argument. So that's why I questioned if we should always include `convert orientation` instead of being part of the attributes in the overrides settings

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -114,47 +120,95 @@ def process(self, instance):
...
             cls, instance
         )
 
-        visible = publish_attributes.get("convert_orientation", cls.convert_orientation)
+        defs = []
+
+        visible = (
+            publish_attributes.get("convert_orientation", cls.convert_orientation)
+            | cls.convert_orientation
+        )
```
BigRoy: Why the `| cls.convert_orientation` here? Seems wrong?

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -114,47 +120,95 @@ def process(self, instance):
...
         }
-                
-        return [
-            BoolDef("convert_orientation",
+
+        overrides = publish_attributes.get("overrides", cls.overrides)
+        if not overrides:
+            return defs
+
+        override_defs = OrderedDict({
```
BigRoy: I feel like we don't need the `OrderedDict` here - Blender 2.83 LTS is the first version that shipped with Py3.7 which has a guaranteed insertion ordering for `dict`. Assuming we don't support Blender even further back I think we can drop the OrderedDict here.

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -114,47 +120,95 @@ def process(self, instance):
...
             "Z": "Z",
             "NEGATIVE_X": "-X",
             "NEGATIVE_Y": "-Y",
             "NEGATIVE_Z": "-Z",
         }
-                
-        return [
-            BoolDef("convert_orientation",
+
+        overrides = publish_attributes.get("overrides", cls.overrides)
```
BigRoy: The `overrides` is a Settings thing - and should never be coming from the publish attributes themselves, right?

```suggestion
        overrides: set[str] = set(cls.overrides)
```

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -15,12 +16,20 @@ class ExtractUSD(plugin.BlenderExtractor,
...
+    # Settings
     convert_orientation = False
+    export_animation = False
+    export_hair = False
+    export_uvmaps = True
+    export_normals = True
+    export_materials = True
+    use_instancing = True
+
+    overrides = []
```
BigRoy: Type hint here may be nice for our future selves
```suggestion
    overrides: list[str] = []
```

client/ayon_blender/plugins/publish/extract_usd.py
moonyuet: But if I set `visible = publish_attributes.get("convert_orientation", cls.convert_orientation)`, I would only get False value for both `cls.convert_orientation` set to `True` or `False`(you know `cls.convert_orientation` can be driven by ayon settings), we somehow included that particular `convert_orientation` as key in the publish_attributes and explicitly set it to False.

BigRoy: So you're saying it doesn't update live to the state of the attribute when it's visible in the UI. That's correct, you'll need to "track changes" yourself. Similar to: https://github.com/ynput/ayon-core/blob/e9689dd54c2fdb578ef2f6e30c48dca01e3f54c8/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py#L668-L701

moonyuet: It doesn't work at all for this case as the `convert_orientation` is already in the callback.

BigRoy: It works fine for me without the `| cls.convert_orientation`. How can I reproduce?

```python
        visible = publish_attributes.get(
            "convert_orientation",
            cls.convert_orientation
        )
```

moonyuet: > It works fine for me without the `| cls.convert_orientation`. How can I reproduce?
> 
> ```python
>         visible = publish_attributes.get(
>             "convert_orientation",
>             cls.convert_orientation
>         )
> ```

It could have been the issue of development mode then:(

client/ayon_blender/plugins/publish/extract_usd.py
```diff
@@ -114,47 +120,99 @@ def process(self, instance):
...
+                "export_mesh_colors",
+                label="Export Mesh Colors",
+                tooltip="Whether to export mesh color data or not.",
+                default=cls.export_mesh_colors),
+            "use_instancing": BoolDef(
+                "use_instancing",
+                label="Instancing",
+                tooltip="Whether to use USD instancing for duplicated objects or not.",
+                default=cls.use_instancing),
+        }
```
BigRoy, line 220: Cosmetics + add tooltip to Forward axis and Up axis.
```suggestion
        override_defs = {
            "convert_orientation": BoolDef(
                "convert_orientation",
                label="Convert Orientation",
                tooltip="Convert orientation axis to a different convention"
                        " to match other applications.",
                default=cls.convert_orientation,
            ),
            "forward_axis": EnumDef(
                "forward_axis",
                label="Forward Axis",
                items=orientation_axes,
                default="Z",
                tooltip="Forward Axis for orientation conversion.",
                visible=visible,
            ),
            "up_axis": EnumDef(
                "up_axis",
                label="Up Axis",
                items=orientation_axes,
                default="Y",
                tooltip="Up Axis for orientation conversion.",
                visible=visible,
            ),
            "export_animation": BoolDef(
                "export_animation",
                label="Export Animation",
                tooltip="Whether to export animation data or not.",
                default=cls.export_animation,
            ),
            "export_hair": BoolDef(
                "export_hair",
                label="Export Hair",
                tooltip="Whether to export hair/fur systems or not.",
                default=cls.export_hair,
            ),
            "export_uvmaps": BoolDef(
                "export_uvmaps",
                label="Export UV Maps",
                tooltip="Whether to export UV map data or not.",
                default=cls.export_uvmaps,
            ),
            "export_normals": BoolDef(
                "export_normals",
                label="Export Normals",
                tooltip="Whether to export normal data or not.",
                default=cls.export_normals,
            ),
            "export_materials": BoolDef(
                "export_materials",
                label="Export Materials",
                tooltip="Whether to export material assignments and data or"
                        " not.",
                default=cls.export_materials,
            ),
            "export_mesh_colors": BoolDef(
                "export_mesh_colors",
                label="Export Mesh Colors",
                tooltip="Whether to export mesh color data or not.",
                default=cls.export_mesh_colors,
            ),
            "use_instancing": BoolDef(
                "use_instancing",
                label="Instancing",
                tooltip="Whether to use USD instancing for duplicated objects"
                        " or not.",
                default=cls.use_instancing,
            ),
        }
```

client/ayon_blender/plugins/publish/extract_usd.py
moonyuet: Fixed in [bffe111](https://github.com/ynput/ayon-blender/pull/295/commits/bffe1116f1520102eb6034f61efbfe1d6d85c6b9)

BigRoy (approved):
This seems to work. 

The only caveat is that with the Export Animation toggle there's no way to set the frame range for the instance, because the USD creator in Blender apparently exposes no frame range options. That's something we should add - but probably not as part of this PR.

PR #306, opened 2026-06-11
Workfile template: Check on window is None when opening template

### LINKED ISSUE #307: YN-0824: templated_workfile: windows return None when opening template is performed
### Please describe the issue you have encountered?

 

When opening template is executed from AYON plugins, it errors out as following:

File "D:\ayon-addon_template\ayon-blender\client\ayon_blender\api\ops.py", line 477, in execute

raise RuntimeError("Failed to open template - window is None")

RuntimeError: Failed to open template - window is None

05:23.750 bpy.rna | ERROR Python script error in WM_OT_ayon_open_template.execute

05:23.750 operator | ERROR Python: Traceback (most recent call last):

| File "D:\ayon-addon_template\ayon-blender\client\ayon_blender\api\ops.py", line 477, in execute

| raise RuntimeError("Failed to open template - window is None")

| RuntimeError: Failed to open template - window is None

&nbsp;

### How to replicate the issue?

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0824]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=e7d3a2b065a211f1a5375736d53c8998)

moonyuet:
## Changelog Description
This PR is to ensure opening template won't hit error and still performs as usual along with other plugins. 

fix https://github.com/ynput/ayon-blender/issues/307

## Testing notes:
1. Launch Blender
2. Templated Workfile -> Open Template
3. The window-related error from ops.py should be gone.

### DIFF
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -471,6 +471,8 @@ class OpenTemplate(LaunchQtApp):
     def execute(self, context):
         from .workfile_template_builder import open_template
         window = open_template()
+        if window is None:
+            return {"FINISHED"}
         BlenderApplication.store_window(self.bl_idname, window)
         self._window = window
         return super().execute(context)
--- a/client/ayon_blender/api/ops.py
+++ b/client/ayon_blender/api/ops.py
@@ -217,7 +217,7 @@ def _process_app_events() -> Optional[float]:
 class LaunchQtApp(bpy.types.Operator):
     """A Base class for operators to launch a Qt app."""
 
-    _window = Union[QtWidgets.QDialog, ModuleType]
+    _window: Optional[Union[QtWidgets.QDialog, ModuleType]] = None
     _tool_name: str = None
     _init_args: Optional[List] = list()
     _init_kwargs: Optional[Dict] = dict()
@@ -250,7 +250,9 @@ class LaunchQtApp(bpy.types.Operator):
 
         if self._tool_name is None:
             if self._window is None:
-                raise AttributeError("`self._window` is not set.")
+                self._window = BlenderApplication.get_window(self.bl_idname)
+                if self._window is None:
+                    raise AttributeError("`self._window` is not set.")
 
         else:
             window = BlenderApplication.get_window(self.bl_idname)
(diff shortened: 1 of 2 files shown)


client/ayon_blender/api/ops.py
```diff
@@ -471,6 +471,8 @@ class OpenTemplate(LaunchQtApp):
     def execute(self, context):
         from .workfile_template_builder import open_template
         window = open_template()
+        if window is None:
+            return {"FINISHED"}
```
BigRoy: This is not the right fix - it may now prompt the dialog on `open_template_ui` but we're not persisting that dialog anywhere if the main window is `None` and hence it should be that Qt kills that dialog instantly. (or maybe it doesn't if it's a modal dialog, but if it's a modal dialog we probably don't need to do all this "main window" passing through to begin with?

Please investigate further what's needed.

moonyuet: It only happened to this `open_template()` function but not the rest of the plugins.

BigRoy: Please see implementation of `open_template` and compare with the others. It doesn't actually return the window it pops up - but it sounds like perhaps it should?

It's because it generates a modal QMessageBox that will be closed before the code continues, as such `open_template` always returns `None`.

So:
- I'm clueless as to why `open_template` even takes `*args, **kwargs` because it's not used anywhere?
- The `kwargs.get("main_window")` there will always be `None` because I don't see any code that passes `main_window` to `open_template`?
- As such, this would always just `return {"FINISHED"}` which I think seems correct then.

In short - a lot should be thrown away, because it's just not used and incorrect.

Also, the same error should appear in Blender 4 as well - it's not B5 specific I'd say.
Unless the actual issue in B5 is that you can't assign `self._window` on these classes - if that's the case, I'd say that the `CreatePlaceholder` call etc. would also error (even though it _may_ show the UI, but an error may still occur)

moonyuet: I believe it only happened on open_template? See the video: 

https://github.com/user-attachments/assets/6ca53ea4-ef1c-41db-9243-f4e74749b18e

BigRoy (approved):

PR #309, opened 2026-06-12
First workfile creation with templated workfile build

### LINKED ISSUE #308: YN-0825: workfile_build: automatically open and save the first workfile

 

What is "Create First Version" doing, I have presumed based on the name that it would if no workfile was found use the template to create the first workfile version

### Additional context:

_This issue was automatically created from AYON ticket [YN-0825]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=a725d64065a511f18b5b5977fb63f9a2)

moonyuet:
## Changelog Description
This PR is to add support for automatically creating the first workfile if templated workfile build in addon setting is provided.

Resolves https://github.com/ynput/ayon-blender/issues/308

## Testing notes:
1. Set up your path for your template and set `create first workfile` to True in `ayon+settings://blender/templated_workfile_build/profiles`
2. Launch Blender with the task which has not yet created any workfile
3. It will automatically open and save the template file as the first workfile

### DIFF
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -28,9 +28,6 @@ from ayon_core.pipeline import (
 from ayon_core.pipeline.context_tools import (
     get_current_task_entity
 )
-from ayon_core.pipeline.workfile.workfile_template_builder import (
-    TemplateProfileNotFound
-)
 from ayon_core.lib import (
     Logger,
     register_event_callback,
@@ -417,11 +414,6 @@ def _create_first_workfile_from_template_timer() -> Optional[float]:
         )
         _is_opening_workfile_template = True
         create_first_workfile_from_template()
-
-    except TemplateProfileNotFound:
-        # to avoid looping of the callback, remove it!
-        log.warning("Template profile not found. Skipping...")
-        _is_opening_workfile_template = False
 
     except Exception:
         log.warning(
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -420,7 +420,6 @@ def _create_first_workfile_from_template_timer() -> Optional[float]:
             "Failed to create first workfile from template on new file.",
             exc_info=True,
         )
-        _is_opening_workfile_template = False
 
     finally:
         _is_opening_workfile_template = False
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -407,11 +407,11 @@ def _create_first_workfile_from_template_deferred() -> Optional[float]:
     if not window_manager or not scene:
         return 0.1
 
+    global _is_opening_workfile_template
     try:
         from .workfile_template_builder import (
             create_first_workfile_from_template,
         )
-        global _is_opening_workfile_template
         _is_opening_workfile_template = True
         create_first_workfile_from_template()
     except Exception:
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -391,6 +391,11 @@ def on_new():
 
     set_unit_scale_from_settings(blender_settings=settings)
 
+    if not IS_HEADLESS:
+        from .workfile_template_builder import create_first_workfile_from_template
+        create_first_workfile_from_template()
+
+
 
 def on_open():
     project = os.environ.get("AYON_PROJECT_NAME")
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -396,7 +396,6 @@ def on_new():
         create_first_workfile_from_template()
 
 
-
 def on_open():
     project = os.environ.get("AYON_PROJECT_NAME")
     settings = get_blender_settings(project)
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -70,6 +70,7 @@ ORIGINAL_EXCEPTHOOK = sys.excepthook
 
 
 log = Logger.get_logger(__name__)
+_is_opening_template = False
 
 
 class BlenderHost(HostBase, IWorkfileHost, IPublishHost, ILoadHost):
@@ -410,6 +411,8 @@ def _create_first_workfile_from_template_deferred() -> Optional[float]:
         from .workfile_template_builder import (
             create_first_workfile_from_template,
         )
+        global _is_opening_template
+        _is_opening_template = True
         create_first_workfile_from_template()
     except Exception:
         log.warning(
@@ -511,7 +514,8 @@ def _on_load_post(*args):
         # Likely this was an open operation since it has a filepath
         emit_event("open")
     else:
-        emit_event("new")
+        if not _is_opening_template:
+            emit_event("new")
 
     ops.OpenFileCacher.post_load()
 
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -28,6 +28,9 @@ from ayon_core.pipeline import (
 from ayon_core.pipeline.context_tools import (
     get_current_task_entity
 )
+from ayon_core.pipeline.workfile.workfile_template_builder import (
+    TemplateProfileNotFound
+)
 from ayon_core.lib import (
     Logger,
     register_event_callback,
@@ -414,11 +417,19 @@ def _create_first_workfile_from_template_timer() -> Optional[float]:
         )
         _is_opening_workfile_template = True
         create_first_workfile_from_template()
+
+    except TemplateProfileNotFound:
+        # to avoid looping of the callback, remove it!
+        log.warning("Template profile not found. Skipping...")
+        _is_opening_workfile_template = False
+
     except Exception:
         log.warning(
             "Failed to create first workfile from template on new file.",
             exc_info=True,
         )
+        _is_opening_workfile_template = False
+
     finally:
         _is_opening_workfile_template = False
 
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -70,7 +70,7 @@ ORIGINAL_EXCEPTHOOK = sys.excepthook
 
 
 log = Logger.get_logger(__name__)
-_is_opening_template = False
+_is_opening_workfile_template = False
 
 
 class BlenderHost(HostBase, IWorkfileHost, IPublishHost, ILoadHost):
@@ -411,14 +411,17 @@ def _create_first_workfile_from_template_deferred() -> Optional[float]:
         from .workfile_template_builder import (
             create_first_workfile_from_template,
         )
-        global _is_opening_template
-        _is_opening_template = True
+        global _is_opening_workfile_template
+        _is_opening_workfile_template = True
         create_first_workfile_from_template()
     except Exception:
         log.warning(
             "Failed to create first workfile from template on new file.",
             exc_info=True,
         )
+    finally:
+        _is_opening_workfile_template = False
+
     return None
 
 
@@ -514,7 +517,7 @@ def _on_load_post(*args):
         # Likely this was an open operation since it has a filepath
         emit_event("open")
     else:
-        if not _is_opening_template:
+        if not _is_opening_workfile_template:
             emit_event("new")
 
     ops.OpenFileCacher.post_load()
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -396,7 +396,7 @@ def on_new():
         _deferred_create_first_workfile_from_template()
 
 
-def _create_first_workfile_from_template_deferred() -> Optional[float]:
+def _create_first_workfile_from_template_timer() -> Optional[float]:
     """Build first workfile from template after Blender startup settles.
 
     Running this during load_post/homefile initialization can be unstable,
@@ -428,12 +428,12 @@ def _create_first_workfile_from_template_deferred() -> Optional[float]:
 def _deferred_create_first_workfile_from_template() -> None:
     """Schedule deferred first workfile template creation once."""
     if bpy.app.timers.is_registered(
-        _create_first_workfile_from_template_deferred
+        _create_first_workfile_from_template_timer
     ):
         return
 
     bpy.app.timers.register(
-        _create_first_workfile_from_template_deferred,
+        _create_first_workfile_from_template_timer,
         first_interval=0.1,
     )
 
--- a/client/ayon_blender/api/pipeline.py
+++ b/client/ayon_blender/api/pipeline.py
@@ -392,8 +392,44 @@ def on_new():
     set_unit_scale_from_settings(blender_settings=settings)
 
     if not IS_HEADLESS:
-        from .workfile_template_builder import create_first_workfile_from_template
+        _deferred_create_first_workfile_from_template()
+
+
+def _create_first_workfile_from_template_deferred() -> Optional[float]:
+    """Build first workfile from template after Blender startup settles.
+
+    Running this during load_post/homefile initialization can be unstable,
+    so defer execution and retry until context is ready.
+    """
+    window_manager = getattr(bpy.context, "window_manager", None)
+    scene = getattr(bpy.context, "scene", None)
+    if not window_manager or not scene:
+        return 0.1
+
+    try:
+        from .workfile_template_builder import (
+            create_first_workfile_from_template,
+        )
         create_first_workfile_from_template()
+    except Exception:
+        log.warning(
+            "Failed to create first workfile from template on new file.",
+            exc_info=True,
+        )
+    return None
+
+
+def _deferred_create_first_workfile_from_template() -> None:
+    """Schedule deferred first workfile template creation once."""
+    if bpy.app.timers.is_registered(
+        _create_first_workfile_from_template_deferred
+    ):
+        return
+
+    bpy.app.timers.register(
+        _create_first_workfile_from_template_deferred,
+        first_interval=0.1,
+    )
 
 
 def on_open():
--- a/client/ayon_blender/api/workfile_template_builder.py
+++ b/client/ayon_blender/api/workfile_template_builder.py
@@ -1,6 +1,5 @@
 """Blender workfile template builder implementation"""
 import bpy
-import os
 
 import itertools
 from ayon_core.pipeline import registered_host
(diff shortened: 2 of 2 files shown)


BigRoy (dismissed):
There's one big issue here unfortunately, and that is:

### It doesn't _open_ the file, but appends

- Blender by default already comes with a starting file (the cube in the scene)
- Template workfile build _appends_ the scene file to the current scene and then builds, meaning you now have the cube AND your template scene merged. Which I'd say is very confusing!
  - Note: This startup file can also be customized via File > Defaults > Save Startup File

I personally think this is a design issue in how workfile template builder has been designed - that it 'imports' the template instead of opening, because there's just so much data that can't be imported in many DCCs but they can be set in a scene file, as such opening is a much better standard way of getting the scene file. I recall the only reason we imported instead of opened - was that we didn't want the filepath of the current scene to be set to the template (because you don't want to accidentally work in it) even though majority of DCCs do support to unset a filepath, or we could always save it in temp if need be.

I would personally argue it should not merge with the startup file but it should _act_ as your startup file - so it should really open it or something equivalent to it. @LiborBatek can you confirm that makes sense to you to?

---

Other than that, this works like a charm!

LiborBatek:
Yeah def true and fully agreed! We should do `File Open` not `File Append` really....

And funny fact, it worked exactly like this before the Template Builder was fully introduced to Blender integration...simply configuring a path to a blend scene which event. got opened when no workfile existed at the time...

A lot of users also been craving for such simple functionality tbh.

I can imagine having 2 options and introduce a separate settings in the `ayon-blender`.... same goes for `maya` host etc...

@BigRoy  so what we should do then?

LiborBatek (commented):
The essentials works but via `File Append` function which has some undesired effects like not bringing full scene setup (e.g. HDRI setting, Render Output attributes like file format etc)

So we should really go with `File > Open` action when using such template blend scene...

I will leave it to you guys how to treat both the case above and when full building from template using placeholders been used instead (as mentioned, maybe separate settings for each might be good)

Here is my template scene:

Image

And once "applied" and first workfile been saved (note the differences):

Image

Also worth mentioning that all settings works really well and been tested throughoutl (filters, saving of the first workfile option etc) all good on this front!

Also tried hard coded filepath and tokenized one and all been working nicely. And lastly tested in both `Blender 4.5.x` and `Blender 5.1`

Image

moonyuet:
I just committed [16cdbba](https://github.com/ynput/ayon-blender/pull/309/commits/16cdbba2848140b40bc408bea7700d52a5a0b3f0) from appending the blend file to opening the blend file. From now, when we build workfile from template, it will open the file from template save the new workfile if the `create first workfil` enabled. This also means that when we build the workfile by using the template workfile build in AYON menu, it will open scene from template (not really import) too.

client/ayon_blender/api/pipeline.py
```diff
@@ -391,6 +391,11 @@ def on_new():
 
     set_unit_scale_from_settings(blender_settings=settings)
 
+    if not IS_HEADLESS:
+        from .workfile_template_builder import create_first_workfile_from_template
+        create_first_workfile_from_template()
+
+
```
BigRoy, line 442:
```suggestion
```

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -42,32 +41,9 @@ def import_template(self, path):
         Returns:
             bool: Whether the template was successfully imported or not
         """
-        if bpy.data.collections.get(PLACEHOLDER_SET):
-            raise TemplateAlreadyImported((
-                "Build template already loaded\n"
-                "Clean scene if needed (File > New Scene)"
-            ))
```
BigRoy, line 49: With placeholder set now being irrelevant, can we also remove the other `PLACEHOLDER_SET` logic?

BigRoy (commented):
Looking good - just some small notes from me. Then happy for @LiborBatek to test it!

client/ayon_blender/api/workfile_template_builder.py
moonyuet: Fixed in [b98b9bb](https://github.com/ynput/ayon-blender/pull/309/commits/b98b9bb32fb284a461d91f82ce738f73100895f4)

client/ayon_blender/api/pipeline.py
```diff
@@ -391,6 +391,10 @@ def on_new():
 
     set_unit_scale_from_settings(blender_settings=settings)
 
+    if not IS_HEADLESS:
+        from .workfile_template_builder import create_first_workfile_from_template
+        create_first_workfile_from_template()
```
BigRoy, line 416: What happens on startup if you do not have a template file set? Does it error or just skip it? As far as I can quickly see the `create_first_workfile_from_template` call will not actually 'ignore' it if it no template is set? 🤔 

I'd expect it to not error - and possibly be silent about it anyway.

moonyuet: I have logged the warning message and return False instead. 
Committed in [f9e10fa](https://github.com/ynput/ayon-blender/pull/309/commits/f9e10fa2bea05552f64bf624808d76de94fddb16)

LiborBatek (changes requested):
While it works nicely in `Blender 4.5.3` it crashes on launch while testing with `Blender 5.1`


My scene been 1:1 to the template file which is perfect (blender 4.5.3)

Image

I have also tried to disable the `Create first version` setting to see if it crashes even when disabled but yes...in both cases or rather in any circumstances.

Image

Image

Im also enclosing full crash log from the Blender:

[Blnd_asset_testing_modeling_v001.crash.txt](https://github.com/user-attachments/files/29086714/Blnd_asset_testing_modeling_v001.crash.txt)

LiborBatek (approved):
Now it works for both `Blender 4.5.x` and `Blender 5.x` and all good!

LGTM

LiborBatek:
@moonyuet @BigRoy  just a small side note: when disabling the option `Create first version` meaning the actual workfile using the template file wont be saved (which is correct and expected) and just stays opened...there is a potential risk the user alter and tweak unintentionally the `template file` itself as its basically just opened and active blend scene... so it might be cool to have sort of function when this `Create first version` been disabled it would actually open the template as a copy (not sure if possible to achieve but would be bulletproof then...

here you can see the situation in real:

Screenshot 2026-06-18 141759

client/ayon_blender/api/pipeline.py
BigRoy, line 422: This should now also have a `finally:` clause that would set this back to False `_is_opening_template = False`

moonyuet: Fixed in [5bc3665](https://github.com/ynput/ayon-blender/pull/309/commits/5bc366570254a5bbfe8ea7c2ac7d8fe8923bfc89)

client/ayon_blender/api/pipeline.py
```diff
@@ -391,6 +392,51 @@ def on_new():
...
+    window_manager = getattr(bpy.context, "window_manager", None)
+    scene = getattr(bpy.context, "scene", None)
+    if not window_manager or not scene:
+        return 0.1
+
+    try:
+        from .workfile_template_builder import (
+            create_first_workfile_from_template,
+        )
+        global _is_opening_workfile_template
```
BigRoy: Just to make sure this global actually also applies to the finally clause - and making it clear that it should.
```suggestion

    global _is_opening_workfile_template
    try:
        from .workfile_template_builder import (
            create_first_workfile_from_template,
        )
```

client/ayon_blender/api/pipeline.py
```diff
@@ -391,6 +392,51 @@ def on_new():
 
     set_unit_scale_from_settings(blender_settings=settings)
 
+    if not IS_HEADLESS:
+        _deferred_create_first_workfile_from_template()
+
+
+def _create_first_workfile_from_template_deferred() -> Optional[float]:
```
BigRoy: I wonder if it's clearer if we refer to this as `_timer` instead of `_deferred` since it may explain more why this returns the float or None

BigRoy (approved):
Works for me. @LiborBatek can you give a rundown of it across blender versions now.

LiborBatek (approved):
Now all been working great for both `Blender 4.5.x` and `Blender 5.x`

LGTM

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -42,32 +37,10 @@ def import_template(self, path):
...
-
-        placeholder_collection = bpy.data.collections.new(PLACEHOLDER_SET)
-        bpy.context.scene.collection.children.link(placeholder_collection)
-        filepath = Path(path).resolve()
-        if not filepath.exists():
+        if not os.path.exists(path):
+            self.log.warning(f"Template file not found: {path}")
```
BigRoy: If you have no workfile template set - does this logic still trigger and log a warning? @LiborBatek @moonyuet if so, we should make it so that with no workfile template set it should stop silently.

moonyuet: I got this in my log
```
ayon_core.pipeline.workfile.workfile_template_builder.TemplateLoadFAILED: Template path is not set.
Path need to be set in Blender\Template Workfile Build Settings\Profiles
Initialising WM_OT_ayon_build_workfile_from_template...
Initialising WM_OT_ayon_build_workfile_from_template...
  File "D:\ayon-addon_template\ayon-blender\client\ayon_blender\api\ops.py", line 450, in execute
    build_workfile_template()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "D:\ayon-addon_template\ayon-blender\client\ayon_blender\api\workfile_template_builder.py", line 198, in build_workfile_template
    builder.build_template()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher-1896\ayon_core\pipeline\workfile\workfile_template_builder.py", line 556, in build_template
    preset = self.get_template_preset()
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher-1896\ayon_core\pipeline\workfile\workfile_template_builder.py", line 874, in get_template_preset
    raise TemplateLoadFailed((
    ...<3 lines>...
    ).format(host_name.title()))
ayon_core.pipeline.workfile.workfile_template_builder.TemplateLoadFailed: Template path is not set.
Path need to be set in Blender\Template Workfile Build Settings\Profiles
00:21.422  bpy.rna          | ERROR Python script error in WM_OT_ayon_build_workfile_from_template.execute
00:21.422  operator         | ERROR Python: Traceback (most recent call last):
                            |   File "D:\ayon-addon_template\ayon-blender\client\ayon_blender\api\ops.py", line 450, in execute
                            |     build_workfile_template()
                            |     ~~~~~~~~~~~~~~~~~~~~~~~^^
                            |   File "D:\ayon-addon_template\ayon-blender\client\ayon_blender\api\workfile_template_builder.py", line 198, in build_workfile_template
                            |     builder.build_template()
                            |     ~~~~~~~~~~~~~~~~~~~~~~^^
                            |   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher-1896\ayon_core\pipeline\workfile\workfile_template_builder.py", line 556, in build_template
                            |     preset = self.get_template_preset()
                            |   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher-1896\ayon_core\pipeline\workfile\workfile_template_builder.py", line 874, in get_template_preset
                            |     raise TemplateLoadFailed((
                            |     ...<3 lines>...
                            |     ).format(host_name.title()))
                            | ayon_core.pipeline.workfile.workfile_template_builder.TemplateLoadFailed: Template path is not set.
                            | Path need to be set in Blender\Template Workfile Build Settings\Profiles
```

moonyuet: I committed one more condition check on whether the path is None or not in [5ec37c1](https://github.com/ynput/ayon-blender/pull/309/commits/5ec37c1b6c1ea092b5c1840e181f43772f597823)

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -42,32 +38,13 @@ def import_template(self, path):
...
-                "Build template already loaded\n"
-                "Clean scene if needed (File > New Scene)"
-            ))
-
-        placeholder_collection = bpy.data.collections.new(PLACEHOLDER_SET)
-        bpy.context.scene.collection.children.link(placeholder_collection)
-        filepath = Path(path).resolve()
-        if not filepath.exists():
+        if not path:
+            self.log.warning("No template path provided.")
```
BigRoy: It not being set may just mean they did it intentional? As in, they may not have any profile whatsoever.
```suggestion
            self.log.debug("No workfile template path set.")
```

client/ayon_blender/api/workfile_template_builder.py
BigRoy: Pretty sure the change here doesn't fix this one - because that happens even before it reaches `import_template` right? 🤔

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -42,32 +38,13 @@ def import_template(self, path):
...
-
-        for target_object in data_dst.objects:
-            bpy.context.scene.collection.objects.link(target_object)
-        for target_collection in data_dst.collections:
-            bpy.context.scene.collection.children.link(target_collection)
-
-
-        # update imported sets information
+        if not os.path.exists(path):
+            self.log.warning(f"Template file not found: {path}")
```
BigRoy: Let's phrase a bit more what sort of template we're looking for.
```suggestion
            self.log.warning(f"Workfile template file not found: {path}")
```

client/ayon_blender/api/workfile_template_builder.py
moonyuet: > Pretty sure the change here doesn't fix this one - because that happens even before it reaches `import_template` right? 🤔

Can we merge this PR and fix in core instead? But hold on, I believe it is expected behavior, it is raising the exception from what we have set up from the core addon.

BigRoy: If this will now always error on launch if you don't have any workfile template set - then we should not merge this before it's resolved because it'd be very annoying otherwise.

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -234,10 +187,21 @@ def set_folder_path_for_ayon_instances(folder_path: str) -> None:
...
+
+    except TemplateProfileNotFound:
+        log.warning(
+            "Template profile not found. Skipping..."
+        )
+
+    except TemplateLoadFailed:
+        log.warning(
+            "Template path not set. Skipping..."
+    )
```
BigRoy: I think a load failure - where a profile is present but the path is just wrong or does not exist, etc. should be considered an error though. At that point, once you have a profile.. then you're intentionally enabling it and expecting it to work.

client/ayon_blender/api/pipeline.py
```diff
@@ -391,6 +392,54 @@ def on_new():
...
+        )
+        _is_opening_workfile_template = True
+        create_first_workfile_from_template()
+
+    except Exception:
+        log.warning(
+            "Failed to create first workfile from template on new file.",
+            exc_info=True,
+        )
+        _is_opening_workfile_template = False
```
BigRoy: Isn't this redundant here if it's also in `finally`? As far as I know `finally` is called both in the case of failure/exception and on no exception. That's kind of the point of the `finally` clause here?
```suggestion
```

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -42,34 +40,9 @@ def import_template(self, path):
...
-        for target_object in data_dst.objects:
-            bpy.context.scene.collection.objects.link(target_object)
-        for target_collection in data_dst.collections:
-            bpy.context.scene.collection.children.link(target_collection)
-
-
-        # update imported sets information
+        bpy.ops.wm.read_homefile(filepath=path)
         update_content_on_context_change()
-        return True
```
BigRoy, line 72: This should still `return True` according to the type hint.

BigRoy (approved):
Looks good - just some cosmetic notes remaining.

client/ayon_blender/api/workfile_template_builder.py
moonyuet: I will remove the type hint then

client/ayon_blender/api/workfile_template_builder.py
```diff
@@ -37,8 +37,6 @@ def import_template(self, path):
             path (str): A path to current template (usually given by
             get_template_preset implementation)
 
-        Returns:
-            bool: Whether the template was successfully imported or not
```
BigRoy, line 41: Ah yes, even better - because the parent class also does not return bool. 👍

LiborBatek (changes requested):
So all been working well for both Blender versions `4.5.x` and `5.x`

but there is a funny glitch that whenever using some `template file` for creating the first workfile it auto pulls it on triggering `CTRL+N` or `Create New Scene` so user is basically not allowed to use clean scene/slate...but always pushed to be using that template file...which is not desirable.

Look at the vid recording showcasing it (also there is no difference when enabled/disabled `Create first workfile` option )

https://github.com/user-attachments/assets/764fde3d-fede-4d30-9eb4-0dc2f5278021

BigRoy:
> but there is a funny glitch that whenever using some template file for creating the first workfile it auto pulls it on triggering CTRL+N or Create New Scene so user is basically not allowed to use clean scene/slate...but always pushed to be using that template file...which is not desirable.

This is _actually_ the correct behavior. We just need a setting to allow disabling it. But same happens in Houdini (and Nuke I think?)

moonyuet:
> > but there is a funny glitch that whenever using some template file for creating the first workfile it auto pulls it on triggering CTRL+N or Create New Scene so user is basically not allowed to use clean scene/slate...but always pushed to be using that template file...which is not desirable.
> 
> This is _actually_ the correct behavior. We just need a setting to allow disabling it. But same happens in Houdini (and Nuke I think?)

I believe we have a github issue in Core: https://github.com/ynput/ayon-core/issues/1040

BigRoy:
> I believe we have a github issue in Core: [ynput/ayon-core#1040](https://github.com/ynput/ayon-core/issues/1040)

Yes, that's the one. I was looking for it but couldn't find it quickly - so thanks!

LiborBatek (approved):
So after the conversation with guys Im giving approval as it works in essence!

PR #312, opened 2026-06-24
Implement Collect Colorspace data as target colorspace data

### LINKED ISSUE #165: Implementing collected colorspace data as targetColorspaceData

DCC implementation of https://github.com/ynput/ayon-core/issues/1430

### Suggested implementation?

as root issue DCC implementation of https://github.com/ynput/ayon-core/issues/1430

_No response_

moonyuet:
## Changelog Description
This PR is to make sure the colorspace data explicitly collector for blender. We would use more explicit naming for view transform and display to differentiate scene colorspace data and the source one.

Fix https://github.com/ynput/ayon-blender/issues/165
As Blender does not support reading the path of the ocio config from their API, it doesn't support for source colorspace data.(We can get this through the environment variables though).

## Testing notes:
1. Create Render Instance
2. Tweak the colorspace from the render settings.
3. Publish

### DIFF
--- a/client/ayon_blender/plugins/publish/collect_render.py
+++ b/client/ayon_blender/plugins/publish/collect_render.py
@@ -176,8 +176,8 @@ class CollectBlenderRender(plugin.BlenderInstancePlugin):
 
         return {
             "colorspaceConfig": ocio_path,
-            "colorspaceDisplay": display,
-            "colorspaceView": view,
+            "sceneDisplay": display,
+            "sceneView": view,
             "colorspace": colorspace
         }
 


BigRoy (approved):
If this is the scene display/view that you view with inside blender, but it does not get written into the output image -> then this does looks good. If this display/view is embedded already within the written image, this should technically be `sourceDisplay` and `sourceView`.

So when testing, please confirm colors are as expected @LiborBatek etc.

LiborBatek (approved):
I have published two distinct renders  (each time recreated the render instance from scratch) while having different `View` space ( rec709 and RAW) and the resulting image looks the same when comparing these two versions ( v001 against v002)

I have also tried to examine the metadata via `exiftool` utility but I couldnt find any trace of these in metadata...

So my assumption been that no scene view/display data been written into the EXR file.

Image

moonyuet:
Can we merge @BigRoy ?

PR #315, opened 2026-07-24
enhancement: attach image/video file from disk as workfile review

sidmehraajm:
Add ExtractWorkfileReviewFromFile publish plugin that exposes a 'Review file' file picker on the Workfile instance. The selected image or video is attached to the workfile product as a 'review'-tagged representation and a thumbnail is generated from it, so a review can be published without rendering a camera/playblast. Leaving the field empty is a no-op.

## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

Detailed information of the changes made to the product or service, providing an in-depth description of the updates and enhancements. This can include technical information, code examples and anything else needed for the review of the PR.

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_blender/plugins/publish/extract_workfile_review_from_file.py
+++ b/client/ayon_blender/plugins/publish/extract_workfile_review_from_file.py
@@ -1,0 +1,178 @@
+"""Attach an existing image/video file as review to the workfile publish.
+
+Unlike ``ExtractPlayblast`` (which requires a review instance with a single
+camera and renders the viewport), this plugin lets the artist attach an
+already existing image or video file straight from disk as the review for the
+*workfile* product.
+
+A "Review file" file picker is added to the Workfile instance in the
+publisher. When a file is picked, it is integrated as a ``review`` tagged
+representation on the workfile product and a thumbnail is generated from it by
+the core ``ExtractThumbnailFromSource`` plugin. Leaving the field empty is a
+no-op, so existing workfile publishes are unaffected.
+"""
+
+import os
+import shutil
+
+import pyblish.api
+
+from ayon_core.lib import FileDef
+from ayon_core.pipeline import publish
+from ayon_core.pipeline.publish import KnownPublishError
+
+from ayon_blender.api import plugin
+
+
+# Extensions offered by the file picker and used to decide how to treat the
+# attached file. Keep them lowercase, with the leading dot.
+VIDEO_EXTENSIONS = {
+    ".mov", ".mp4", ".mkv", ".avi", ".m4v", ".webm", ".mxf", ".wmv",
+}
+IMAGE_EXTENSIONS = {
+    ".png", ".jpg", ".jpeg", ".exr", ".tif", ".tiff", ".bmp", ".gif",
+    ".tga", ".dpx",
+}
+REVIEW_EXTENSIONS = sorted(VIDEO_EXTENSIONS | IMAGE_EXTENSIONS)
+
+
+class ExtractWorkfileReviewFromFile(
+    plugin.BlenderExtractor,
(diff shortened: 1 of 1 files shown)


BigRoy:
Admittedly, I think this has no reason of being inside Blender - it may very well apply nearly as much across most integrations (if not all) which makes me think whether there's any reason to not move something akin to this into AYON-core. It applies to a much wider use case - and it's not blender-specific?

What type of videos are you attaching? Where do they come from? Are they manually made from the scene-file? Because if so, isn't the end goal instead that we automate that generation?

Also, with that said - I feel this is also a fine addition to just a custom studio addon which makes me wonder whether there's actually a place for this out of the box in AYON. Seems rather specific to a preference of doing this as well?

sidmehraajm:
Fair points, and I mostly agree on the structure.

You're right that this isn't specific to Blender. It has already been transferred almost word for word to Maya and Houdini, which supports your argument. The only parts tied to a specific host are the import, the base extractor class, and the hosts/toggle defaults. So yes, it makes sense for it to exist in ayon-core as a single optional extractor that is turned off by default, which every host can use, instead of having multiple copies. I can refactor it that way if there’s interest. The host add-ons would just need to set the base class and the route_through_extract_review default.

Regarding the videos and their source: they aren't playblasts or anything we can automatically create from the scene. The artist captures wireframe images, screenshots, reference photos, various elements, and combines them in DaVinci (or similar software) into a single review video, which they then export. So, the content is curated by a person. It consists of merged references, annotations, and multiple sources, which is exactly why automating the generation isn’t the goal. The finished clip is what we deliver; there’s nothing in the scene that would recreate it.

The specific issue it addresses: right now, to attach that clip as the reviewable, the artist must open AYON in Chrome, find the right product and version, and manually drag the file onto it. This process is cumbersome and prone to mistakes, as it’s easy to attach it to the wrong version. This solution allows them to simply point to the file during publishing and move on.

BigRoy:
> You're right that this isn't specific to Blender. It has already been transferred almost word for word to Maya and Houdini, which supports your argument. The only parts tied to a specific host are the import, the base extractor class, and the hosts/toggle defaults. So yes, it makes sense for it to exist in ayon-core as a single optional extractor that is turned off by default, which every host can use, instead of having multiple copies. I can refactor it that way if there’s interest. The host add-ons would just need to set the base class and the route_through_extract_review default.

You don't need to inherit from host-specific plug-ins, this can really be in core. It doesn't care any blender specific logic.

> Regarding the videos and their source: they aren't playblasts or anything we can automatically create from the scene. The artist captures wireframe images, screenshots, reference photos, various elements, and combines them in DaVinci (or similar software) into a single review video, which they then export. So, the content is curated by a person. It consists of merged references, annotations, and multiple sources, which is exactly why automating the generation isn’t the goal. The finished clip is what we deliver; there’s nothing in the scene that would recreate i

Sounds convoluted admittedly. I feel like you'd almost want to do the reverse and publish the video from DaVinci and attach the model at that point. hehe.

> The specific issue it addresses: right now, to attach that clip as the reviewable, the artist must open AYON in Chrome, find the right product and version, and manually drag the file onto it. This process is cumbersome and prone to mistakes, as it’s easy to attach it to the wrong version. This solution allows them to simply point to the file during publishing and move on.

Yup, waiting for it to finish to _then_ do it manually sounds annoying. I really feel like there's a better strategy here to automate this because it still sounds prone to human error and whatnot. Either way, this really should be in core if it's agnostic.

sidmehraajm:
cool, I can add it in core

PR #316, opened 2026-07-27
Blender 5.2+ compatibility fixes

BigRoy:
## Changelog Description

Blender 5.2 should work correctly within AYON.

1. Avoid error:
```python

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_1.1.6\ayon_blender\api\ops.py", line 126, in execute

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_1.1.6\ayon_blender\plugins\load\load_blend.py", line 237, in exec_update
    self.exec_remove(container)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\blender_1.1.6\ayon_blender\plugins\load\load_blend.py", line 321, in exec_remove
    getattr(bpy.data, attr).remove(data)

AttributeError: bpy_prop_collection: attribute "remove" not found
```

2. Also fix error with `BLENDER_EEVEE_NEXT` found found in render engine enum because it's renamed back to `BLENDER_EEVEE`

Please do more testing an report any issues.
Create issues for those things that we need to add additional support for in Blender 5.2+ (e.g. new features)

## Testing notes:

1. Blender 5.2 should work as intended.

### DIFF
--- a/client/ayon_blender/api/render_lib.py
+++ b/client/ayon_blender/api/render_lib.py
@@ -138,7 +138,7 @@ def set_render_passes(settings, renderer, view_layers):
         aov_list = base_aov_list.union(existing_aov_list)
         aov_list_combined.update(aov_list)
 
-        if renderer == "BLENDER_EEVEE":
+        if _is_legacy_eevee_renderer(renderer):
             # Eevee exclusive passes
             aov_options = get_aov_options(renderer)
             eevee_attrs: set[str] = {
@@ -178,6 +178,29 @@ def set_render_passes(settings, renderer, view_layers):
     return list(aov_list_combined), custom_passes
 
 
+def _is_legacy_eevee_renderer(renderer):
+    """Return whether renderer is Eevee in Blender lower than 4.2.
+
+    Note:
+      In Blender <4.2 'BLENDER_EEVEE' represents Eevee renderer.
+      In Blender 4.2-5.1 'BLENDER_EEVEE_NEXT' represents Eevee renderer.
+      In Blender >5.2 'BLENDER_EEVEE' represents Eevee renderer.
+
+    Args:
+        renderer (str): Renderer name.
+
+    Returns:
+        bool: Whether renderer is Eevee in Blender version lower than 4.2.
+    """
+    if renderer != "BLENDER_EEVEE":
+        return False
+
+    ver_major, ver_minor, _ = lib.get_blender_version()
+    if (ver_major, ver_minor) < (4, 2):
+        return True
+    return False
+
+
 def get_aov_options(renderer: str) -> dict[str, str]:
     """Return the available AOV options based on the renderer name."""
     aov_options = {
@@ -196,7 +219,7 @@ def get_aov_options(renderer: str) -> dict[str, str]:
         "cryptomatte_material": "use_pass_cryptomatte_material",
         "cryptomatte_asset": "use_pass_cryptomatte_asset",
     }
-    if renderer == "BLENDER_EEVEE":
+    if _is_legacy_eevee_renderer(renderer):
         eevee_options = {
             "shadow": "use_pass_shadow",
             "volume_light": "use_pass_volume_direct",
@@ -233,7 +256,7 @@ def existing_aov_options(
 ) -> list[str]:
     aov_list = []
     aov_options = get_aov_options(renderer)
-    if renderer == "BLENDER_EEVEE":
+    if _is_legacy_eevee_renderer(renderer):
         eevee_attrs = ["use_pass_shadow", "cryptomatte_accurate"]
         for pass_name, attr in aov_options.items():
             target = view_layer.eevee if attr in eevee_attrs else view_layer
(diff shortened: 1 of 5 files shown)


client/ayon_blender/plugins/load/load_blend.py
```diff
@@ -312,12 +305,12 @@ def exec_remove(self, container: dict) -> bool:
...
-        for attr in attrs:
-            for data in getattr(bpy.data, attr):
-                if data in members:
-                    # Skip the asset group
-                    if data == asset_group:
+        if members:
+            for _, attr in iter_bpy_prop_collection_idprop():
+                # make a list copy because we remove members as we iterate
+                for data in list(attr):
+                    if data not in members or data == asset_group:
```
moonyuet, line 312: It's quite confusing why data is equal to asset_group,  `members = asset_group.get(AYON_PROPERTY).get("members", [])` in line 298 shows that members should be the attributes from the dictionary of AYON_PROPERTY which is lasted inside the asset group. It should be the first condition which makes sense.
Is it some magical moment with the new attribute's introduction?
```suggestion
                    if data not in members:

**Review by moonyuet — COMMENTED** [review]
Tested Blender 5.1 & 5.2. Looks good
I have just one comment and question above and otherwise it's good to go.

File: client/ayon_blender/plugins/load/load_blend.py (hunk 1 above)
Comment (BigRoy) [reply client/ayon_blender/plugins/load/load_blend.py:312]: Sorry, what? :) 

It's the same check as before, but written differently.

Matching old code:
```
if data in members:
    # Skip the asset group
    if data == asset_group:
        continue
```

`data` is an entry of the value in the attributes, where the attributes are e.g. [BlendData.collections](https://docs.blender.org/api/current/bpy.types.BlendData.html#bpy.types.BlendData.collections), BlendData.images, etc. As such, each "data" is one collection or one image, or any of the other single data blocks of Blender. Meaning that as we iterate these datablocks, we can also pass the node itself.

If by any chance, the node itself is also in members (I wouldn't know why admittedly?) we skip it here. I presume the node itself CAN be linked to its own property and that's where the original if branch came from to avoid issues with deleting the node itself early?

Anyway, logic isn't new.

File: client/ayon_blender/api/render_lib.py [hunk 2]
```diff
@@ -178,6 +178,29 @@ def set_render_passes(settings, renderer, view_layers):
...
+    Returns:
+        bool: Whether renderer is Eevee in Blender version lower than 4.2.
+    """
+    if renderer != "BLENDER_EEVEE":
+        return False
+
+    ver_major, ver_minor, _ = lib.get_blender_version()
+    if (ver_major, ver_minor) < (4, 2):
+        return True
+    return False
```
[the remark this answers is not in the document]
BigRoy: I was actually quite surprised by this too @moonyuet 

Looking at the old code - this was also true for 4.2+ previously so the AOV setup for eevee only ever worked for eevee specific attributes in Blender <4.2 looking at the code.

@LiborBatek @moonyuet 

Do we care? I have a feeling we'll want to drop this code at some point anyway.

client/ayon_blender/plugins/load/load_blend.py
moonyuet: > Sorry, what? :)
> 
> It's the same check as before, but written differently.
> 
> Matching old code:
> 
> ```
> if data in members:
>     # Skip the asset group
>     if data == asset_group:
>         continue
> ```

I see. 
But without the second logic is still removing the content as expected(and no errors, I assume there is some edge case behind).
Although I dont really see why the data from AYON PROPERTY would be lasted in the collections as separate data from AYON_PROPERTY, we can keep it there for some edge cases. There are always some reasons behind those lasted logics.

moonyuet (approved):

client/ayon_blender/plugins/load/load_blend.py
BigRoy: It's always in `bpy.data` (because that has everything the workfile). The question is, what is the edge case that the collection itself is also in this property on the instance - which I'm not sure how one would reproduce that.

It's been there for over two years: https://github.com/ynput/ayon-blender/blame/0f22d515c37ac16c54282bebf410b2a1534b1724/client/ayon_blender/plugins/load/load_blend.py#L319

It was there in OpenPype: https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/blender/plugins/load/load_blend.py#L279

It was introduced by Simone here:
https://github.com/ynput/OpenPype/commit/3d21df1fc5e24e4009d949f5c7172c26660fb5eb

There should be no harm in having the check in there, but I'm with you on this that it makes little sense in practice. Since it doesn't do harm but it _may_ fix something I'm ok with keeping it for now.

LiborBatek (approved):
I have tested both managing products when load appended and linked too, both were working well when setting versions / removing them.

Also tested in both legacy `blender 4.5.3` and `blender 5.2` again working well in both cases...

Also the issue with `Create Render Setup` and Eevee is working now for both blender versions... good!

I havent spotted any issues and LGTM!

MilaKudr:
I tested loading addons, working good.
image

